PR #2181, opened 2026-05-17
chore: refresh npm lockfile and drop stale scheduled_tasks.lock

dsj1984:
Routine lockfile bump (no functional dep changes) plus removal of the stale `.claude/scheduled_tasks.lock` artifact picked up after a session.


dsj1984:
Closing without merge: the lockfile bump generated on Windows is missing Linux-only optional deps (@emnapi/core, @emnapi/runtime), so npm ci fails on Ubuntu runners. The drift was incidental.

PR #2877, opened 2026-05-22
Add gh-exec default timeout config knob + classify GhExecTimeoutError as transient (#2860)

### LINKED ISSUE #2860: Add gh-exec default timeout config knob + classify GhExecTimeoutError as transient
# Add gh-exec default timeout config knob + classify GhExecTimeoutError as transient

> **Reassessed 2026-05-22 — proceed.**
>
> Original preamble offered a "Won't Do" exit ramp if retros were clean after
> #2852 and #2853 landed. Re-review confirms the exit ramp does not apply:
>
> 1. `GhExecTimeoutError`'s message is
>    `"gh-exec: gh <args> exceeded <N>ms"` ([`gh-exec.js`](../../.agents/scripts/lib/gh-exec.js)
>    L357). It contains neither `"timeout"` nor `"timed out"`, so
>    `TRANSIENT_MESSAGES` in
>    [`errors.js`](../../.agents/scripts/providers/github/errors.js) L38-47
>    does not match it. `err.code` and `err.status` are unset on timeout.
>    `classifyGithubError` therefore returns `'permanent'` today, and
>    `withTransientRetry` (from #2852) **re-throws on the first timeout
>    instead of backing off** — exactly the silent failure the original
>    preamble warned about.
> 2. Zero callers in `providers/github/**` pass `timeoutMs`, so
>    `spawnOpts.timeout` is never set and a stalled `gh api` can hang the
>    orchestrator indefinitely. The retry machinery #2852 added cannot
>    fire because the timeout never trips.
>
> Without this Story, #2852's retry helper is half-wired. Proceed.

## Context

`gh-exec.js` accepts a `timeoutMs` parameter per call and will kill the
spawned `gh` process with SIGTERM when the deadline expires, throwing a
`GhExecTimeoutError`. However, two gaps prevent this from being useful
against the retro timeout pattern:

1. **No caller supplies a timeout.** Every provider gateway (`tickets.js`,
   `issues.js`, `comments.js`, etc.) calls `this._gh.api(...)` without
   a `timeoutMs` argument, so every `gh` subprocess can hang indefinitely.
   A TCP stall or a GitHub long-poll on a paginated endpoint will hold a
   retro forever.

2. **`GhExecTimeoutError` is not classified as transient.** The error
   class's message ("gh-exec: gh ... exceeded Nms") matches no keyword in
   `TRANSIENT_MESSAGES`, and `err.status` / `err.code` are unset. Without
   a `'transient'` classification, `withTransientRetry` (landed in Story
   #2852) re-throws immediately on the first timeout rather than backing
   off and retrying — the worst outcome because the retry machinery
   exists but silently won't fire.

The fix has two atomic parts:

- **Part A — Classifier fix.** Teach `classifyGithubError` to recognize
  `GhExecTimeoutError` (preferred: check `err.name === 'GhExecTimeoutError'`
  to avoid a circular import between `providers/github/errors.js` and
  `lib/gh-exec.js`) as `'transient'`. This is ~5 LoC and safe to land on
  its own; it makes the existing retry loop in `addSubIssue` and the new
  `withTransientRetry` wrappers from #2852 all "just work" on gh-exec
  timeouts whenever a timeout is set.

- **Part B — Config knob.** Wire a single `agentSettings.gh.defaultTimeoutMs`
  config key (default `60_000` — 60 s). Resolve it in the provider facade
  constructor and pass it as the default `timeoutMs` for every
  `_gh.api(...)` call, so every gateway inherits a ceiling without code
  changes to each gateway file.

The two parts are logically ordered (A before B) but can be reviewed as
one commit.

Reference files:

- [`gh-exec.js`](../../.agents/scripts/lib/gh-exec.js) — `GhExecTimeoutError`
  class, `timeoutMs` wiring.
- [`errors.js`](../../.agents/scripts/providers/github/errors.js) —
  `classifyGithubError`, `isTransientByCodeOrMessage`.
- Provider facade constructor (locate by tracing the `_gh` instantiation
  — likely `compose.js` or a `GhFacade` class) — where the default
  `timeoutMs` should be threaded.
- [`request-helpers.js`](../../.agents/scripts/providers/github/request-helpers.js)
  — `paginateRest` already wraps each `api()` call with
  `withTransientRetry`; Part A is the fix that makes those retries fire
  on timeout.

## Acceptance Criteria

- [ ] `classifyGithubError` in
      [`errors.js`](../../.agents/scripts/providers/github/errors.js)
      returns `'transient'` for a `GhExecTimeoutError` instance. Prefer
      `err.name === 'GhExecTimeoutError'` (or include `"exceeded "` +
      `"ms"` token-pair match) to avoid a circular import on
      `lib/gh-exec.js`. Confirm that `withTransientRetry` will therefore
      retry gh-exec timeouts up to `maxAttempts` before bubbling.
- [ ] A unit test in
      [`tests/providers/github/errors.test.js`](../../tests/providers/github/errors.test.js)
      asserts `classifyGithubError(new GhExecTimeoutError('...'))` returns
      `'transient'`. Add one negative test: a non-timeout error
      constructed from a plain `Error` is NOT reclassified.
- [ ] `agentSettings.gh.defaultTimeoutMs` is a recognized optional key in
      `.agents/schemas/agentrc.schema.json` (type `integer`, minimum
      `1000`, default `60000` documented in `description`). Existing
      configs without the key continue to work (the default is applied in
      code, not via schema `default`).
- [ ] The provider facade constructor reads
      `agentSettings.gh.defaultTimeoutMs` from the resolved config and
      passes it as a default `timeoutMs` to `GhExec` (or wraps the
      constructed facade so every `api()` call injects `timeoutMs` when
      the caller does not supply one). No individual gateway file needs
      to change.
- [ ] A unit test proves that when `agentSettings.gh.defaultTimeoutMs` is
      set to a small value (e.g., 1 ms) and `gh api` is stubbed to hang,
      the facade propagates the timeout to `GhExec` and the resulting
      `GhExecTimeoutError` bubbles (or, if `withTransientRetry` wraps the
      call, is retried and ultimately bubbles after `maxAttempts`). A
      missing key uses the 60 s default, not hang indefinitely.
- [ ] `npm run lint` and `npm test` pass.

## Out of Scope

- Changing the per-call `timeoutMs` API on `gh-exec.js` itself — it
  already exists and is correct.
- Adding retry logic to any path not already wrapped by
  `withTransientRetry` (that is #2852's domain).
- Touching `addSubIssue` retry — already correct; do not fold it into the
  new helper as part of this Story.
- Raising or lowering `withTransientRetry`'s `maxAttempts`/`capMs`
  defaults. If 60 s × 6 attempts feels wrong after real-world data,
  adjust in a follow-on.
- Any changes to `retro-runner.js` or the orchestration layer — those
  are #2853's domain.

## Notes

- **Module boundary for `GhExecTimeoutError`:** `gh-exec.js` lives under
  `lib/` while `errors.js` lives under `providers/github/`. Prefer
  comparing by `err.name === 'GhExecTimeoutError'` in `errors.js` to
  avoid a circular dependency. Check the existing import graph before
  deciding.
- **Config resolution:** The config resolver is at
  `lib/config-resolver.js`. Search for the pattern used by other
  `agentSettings.*` keys to understand how to thread the new key through
  to the provider layer cleanly.
- **60 s default rationale:** GitHub's own API timeout guidance is ~30 s
  for interactive calls; 60 s gives `paginateRest` one full page of
  headroom before declaring a stall. If a single `gh api` page
  round-trip consistently takes longer than 60 s the problem is
  upstream, not the timeout value.
- **Schema update:** Add the key to
  `.agents/schemas/agentrc.schema.json` under `agentSettings.gh` (create
  the `gh` object if it does not exist). Run `npm run lint` to verify
  the schema is valid JSON.

dsj1984:
Closes #2860

### DIFF
--- a/.agents/schemas/agentrc.schema.json
+++ b/.agents/schemas/agentrc.schema.json
@@ -159,6 +159,11 @@
         "projectNumber": { "type": ["integer", "null"], "minimum": 1 },
         "projectOwner": { "type": ["string", "null"], "minLength": 1 },
         "operatorHandle": { "type": "string", "pattern": "^@.+" },
+        "defaultTimeoutMs": {
+          "type": "integer",
+          "minimum": 1000,
+          "description": "Default `timeoutMs` (in milliseconds) applied to every `gh` subprocess spawned by the GitHub provider facade. Caps any single `gh api` / `gh issue ...` / `gh pr ...` invocation so a stalled TCP socket or long-poll cannot hang an orchestration indefinitely. A `GhExecTimeoutError` from a hit ceiling is classified `transient` and retried by `withTransientRetry`. Recommended floor 1000ms; recommended default 60000 (60s). Omit to use the in-code default of 60000. Story #2860."
+        },
         "branchProtection": { "$ref": "#/$defs/branchProtection" },
         "mergeMethods": { "$ref": "#/$defs/mergeMethods" },
         "notifications": { "$ref": "#/$defs/notifications" }
--- a/.agents/scripts/lib/config-resolver.js
+++ b/.agents/scripts/lib/config-resolver.js
@@ -198,6 +198,7 @@ function buildLegacyShim(blocks) {
             projectNumber: github.projectNumber ?? null,
             projectOwner: github.projectOwner ?? null,
             operatorHandle: github.operatorHandle,
+            defaultTimeoutMs: github.defaultTimeoutMs ?? null,
           },
           notifications: github.notifications,
           worktreeIsolation: delivery.worktreeIsolation,
--- a/.agents/scripts/lib/config-settings-schema.js
+++ b/.agents/scripts/lib/config-settings-schema.js
@@ -192,6 +192,7 @@ const GITHUB_SCHEMA = {
     projectNumber: { type: ['integer', 'null'], minimum: 1 },
     projectOwner: { type: ['string', 'null'], minLength: 1 },
     operatorHandle: { type: 'string', pattern: '^@.+' },
+    defaultTimeoutMs: { type: 'integer', minimum: 1000 },
     branchProtection: BRANCH_PROTECTION_SCHEMA,
     mergeMethods: MERGE_METHODS_SCHEMA,
     notifications: NOTIFICATIONS_SCHEMA,
--- a/.agents/scripts/lib/gh-exec.js
+++ b/.agents/scripts/lib/gh-exec.js
@@ -408,8 +408,15 @@ export function exec({
  * `-X / -f / --paginate / --input -` flag combination.
  *
  * @param {Function} execImpl — exec implementation. Defaults to module exec.
+ * @param {object} [defaultExecOpts] — Default exec options spread into every
+ *   wrapper's invocation (e.g. `{ timeoutMs: 60_000 }`). Per-call options
+ *   override the defaults via spread order. Story #2860.
  */
-export function createGh(execImpl = exec) {
+export function createGh(execImpl = exec, defaultExecOpts = {}) {
+  // Wrap execImpl so every wrapper inherits defaultExecOpts (e.g. timeoutMs)
+  // without having to thread the option through each callsite. Per-call opts
+  // win because they spread after the defaults. Story #2860.
+  const execWithDefaults = (opts) => execImpl({ ...defaultExecOpts, ...opts });
   /**
    * `gh api` wrapper.
    *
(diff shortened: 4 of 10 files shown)


dsj1984:
## 🔬 Code Review — Story #2860

**Comparison**: `main` … `story-2860`
**Provider**: `native`
**Findings**: 2

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 1
- 🟢 Suggestion: 1

### 🟡 Medium Risk (1)

#### 🟡 Size/Volume Warning — `.agents/scripts/lib/config-settings-schema.js` _[maintainability]_

Module `.agents/scripts/lib/config-settings-schema.js` reports a size/volume warning (module 40.8, worst method 40.8).

### 🟢 Suggestion (1)

PR #3366, opened 2026-05-29
single-story-deliver: worktree-isolated edits can silently land in the main checkout (no wrong-tree guard) (#3364)

### LINKED ISSUE #3364: single-story-deliver: worktree-isolated edits can silently land in the main checkout (no wrong-tree guard)
## Summary

`/single-story-deliver` (and `/story-deliver` → `helpers/single-story-deliver`) creates a per-Story worktree via `single-story-init.js` and instructs the agent to `cd` into the returned `workCwd` before editing (Step 0.5; Constraints: "**Always** `cd` into the workCwd before editing"). On Windows this guidance is silently insufficient: `cd <workCwd>` only steers the **Bash** tool's working directory, while the **Edit/Write** tools operate on **absolute paths** and ignore cwd. An agent whose shell is correctly inside the worktree can still edit files under the **main checkout** if it resolves a main-checkout absolute path — the two surfaces silently disagree, and nothing detects the divergence.

Observed during delivery of Story #3358 (PR #3363): the child agent's first round of edits landed on the main checkout instead of the `.worktrees/story-3358/` worktree. It self-corrected by copying the files into the worktree and restoring the main checkout before committing, so the final merge was correct — but only because the agent happened to notice. This is at least the second recorded instance of worktree/main-checkout confusion in this class of workflow.

## Root cause

- `single-story-init.js` correctly returns the worktree path as `workCwd` (`workCwd = ensured.path` at ~line 394–395) — this is **not** a script bug.
- The trap is the mismatch between Bash cwd (which `cd` controls) and the path-based Edit/Write tools (which it does not). The workflow's `cd` instruction cannot enforce edit scope.
- There is **no guard** anywhere in the close path. `single-story-close.js` runs its gates against the worktree only. If edits had gone to the main checkout and the agent had **not** noticed:
  - close would commit an empty / partial worktree (gates pass on the unchanged tree),
  - the main checkout is left with uncommitted changes under tracked paths,
  - the failure is **silent** — no error, the PR just lacks the intended change.

## Impact

Silent wrong-tree edits in standalone Story delivery. Best case (today): the agent self-corrects with a manual copy-and-restore dance. Worst case: a PR merges without the intended change while the main checkout is left dirty, and the operator must notice the empty diff. The blast radius is every `/single-story-deliver` and `/story-deliver` run on a worktree-isolated host.

## Proposed fix (one or more)

1. **Wrong-tree guard in `single-story-close.js`** — before running gates, check the main checkout (`--cwd` repo root) for uncommitted changes under tracked paths. If the main checkout is dirty while the worktree is the active work tree, warn loudly (or abort with a `friction` comment) naming the stray files, so the operator/agent relocates them before commit. Make this the load-bearing detection so it works even when the agent never realizes it edited the wrong tree.
2. **Echo the absolute worktree root for edits** — `single-story-init.js` / the helper Step 0.5 should print the absolute `workCwd` path prominently and instruct the agent that **all Edit/Write paths must be prefixed with that root**, not just `cd` into it. Reduce the trap at the source.
3. **Optional: post-init sanity probe** — a lightweight check the agent can run (or that close runs) comparing `git -C <main-repo> status --porcelain` against `git -C <worktree> status --porcelain` to flag the asymmetry.

Option 1 is the robust, detection-based fix (it catches the failure regardless of agent diligence); option 2 is the cheap preventative ergonomics improvement. They compose well — do both.

## Acceptance

- A standalone-Story close path detects when tracked-path edits have landed in the main checkout instead of the active worktree, and surfaces it (warn + named files at minimum; abort + `friction` comment preferred) rather than silently committing an unchanged worktree.
- The helper workflow makes the worktree-scope requirement explicit for the **path-based** edit tools, not only for the Bash cwd.
- Regression test: simulate a worktree-isolated Story where tracked changes exist only in the main checkout; assert close flags the divergence instead of proceeding to a clean-tree commit.

## Verify

- New unit/contract test for the wrong-tree detection added under `tests/` and passing.
- `npm run test:quick` green.
- Manual: run `/single-story-deliver` on a host with `delivery.worktreeIsolation.enabled = true`, intentionally edit a tracked file in the main checkout, and confirm close flags it.

## Context

- Surfaced delivering Story #3358 / PR #3363 (2026-05-29).
- Related workflow: `.agents/workflows/helpers/single-story-deliver.md` (Step 0.5, Constraints).
- Related script: `.agents/scripts/single-story-init.js` (workCwd handoff), `.agents/scripts/single-story-close.js` (close gates).

dsj1984:
Closes #3364

### DIFF
--- a/.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js
+++ b/.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js
@@ -1,0 +1,225 @@
+/**
+ * phases/wrong-tree-guard.js — detect worktree/main-checkout edit divergence.
+ *
+ * Story #3364 — `/single-story-deliver` (and `/story-deliver`) materializes a
+ * per-Story worktree and instructs the agent to `cd` into it before editing.
+ * On Windows that guidance is silently insufficient: `cd <workCwd>` steers the
+ * Bash tool's working directory, but the path-based Edit/Write tools operate on
+ * absolute paths and ignore cwd. An agent whose shell is correctly inside the
+ * worktree can still resolve a main-checkout absolute path and edit the wrong
+ * tree — the two surfaces disagree and nothing detects it.
+ *
+ * Failure mode this guards: the worktree is the intended work tree, but the
+ * agent's edits landed under the main checkout instead. `single-story-close.js`
+ * runs its gates against the worktree only, so it would commit an unchanged /
+ * partial worktree (gates pass on the clean tree) while leaving the main
+ * checkout dirty — a silent empty-diff PR.
+ *
+ * Detection: when a worktree is the active work tree (it exists on disk and is
+ * distinct from `cwd`), inspect `git -C <mainCheckout> status --porcelain`. Any
+ * **tracked-path** change (modified, staged, deleted, renamed) in the main
+ * checkout is the wrong-tree signal. Untracked files (`??`) are ignored — they
+ * are scratch artifacts, not relocated Story work, and flagging them would
+ * produce false positives on every run.
+ *
+ * The guard runs before the gate chain. On detection it posts a `friction`
+ * structured comment naming the stray files and throws, so the operator/agent
+ * relocates the edits into the worktree before close proceeds. This is the
+ * load-bearing, detection-based fix: it fires regardless of whether the agent
+ * ever realized it edited the wrong tree.
+ */
+
+import path from 'node:path';
+import { postStructuredComment } from '../../ticketing/state.js';
+
+/**
+ * Parse `git status --porcelain` output into structured status entries.
+ *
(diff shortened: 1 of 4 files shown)


dsj1984:
## 🔬 Code Review — Story #3364

**Comparison**: `main` … `story-3364`
**Provider**: `native`
**Findings**: 2

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 1
- 🟢 Suggestion: 1

### 🟡 Medium Risk (1)

#### 🟡 Size/Volume Warning — `.agents/scripts/single-story-close.js` _[maintainability]_

Module `.agents/scripts/single-story-close.js` reports a size/volume warning (module 61.2, worst method 61.2).

### 🟢 Suggestion (1)

PR #3408, opened 2026-06-01
Mandrel: story-close format gate + format-autofix run `biome format .` from worktree cwd — fragile against root ignore globs (#3407)

### LINKED ISSUE #3407: Mandrel: story-close format gate + format-autofix run `biome format .` from worktree cwd — fragile against root ignore globs
## Summary

The story-close pre-merge **format gate** (`.agents/scripts/lib/close-validation.js` → `buildDefaultGates`) and the **format-autofix** step (`.agents/scripts/lib/orchestration/story-close/format-autofix.js`, `format-autofix-scoped.js`) run the formatter as `biome format .` with **`cwd` set to the story worktree** (`.worktrees/story-<id>/`). Running a formatter against `.` from a directory that is itself inside a globally-ignored path is fragile: if the consumer's `biome.json` ignores `**/.worktrees`, the worktree's own absolute path contains a `.worktrees` segment, so `.` matches the ignore glob, **zero files are processed, and biome exits 1** ("No files were processed in the specified paths").

This blocks **every** Story close in an Epic — the gate fails for reasons unrelated to the Story's code.

## Reproduction (athportal, Epic dsj1984/athportal#12)

1. Consumer `biome.json` has `"!**/.worktrees"` in `files.includes` (a reasonable exclusion so the main checkout doesn't format worktree copies).
2. `story-close.js --story <id>` runs the format gate with `cwd=.worktrees/story-<id>`.
3. Gate runs `npx biome format .` → `× No files were processed` → exit 1 → `Pre-merge validation failed at "format"`.
4. `format-autofix` logs `npx biome format --write . exited non-zero (unknown); falling through to the format check gate` for the same reason.

Observed on `biome 2.4.16`, Windows.

## Why the existing knobs don't fix it

- **`project.commands.formatCheck` override** (e.g. `npx biome format --no-errors-on-unmatched .`): only **silences** the exit-1. biome still processes **0 files**, so the gate passes as a **false green** — it verifies nothing. (A prior consumer Story tried exactly this; it masks the bug rather than fixing it.)
- The framework default shipped in `.agents/full-agentrc.json` is the bare `npx biome format .`, which is the failing form.

## Consumer-side root-cause fix (what unblocked us)

Anchor the ignore glob: `"!**/.worktrees"` → `"!.worktrees"`. Root-relative, so it still excludes worktree copies when biome runs from the main checkout, but no longer self-excludes when biome runs from **inside** a worktree. Verified: main checkout = 399 files (worktrees excluded); from inside a worktree = 407 files processed, exit 0; Story close then succeeds with **real** formatting verification.

But this depends on every consumer authoring exactly the right glob shape — the framework should be robust regardless.

## Suggested framework fix (one or more)

1. **Run the format gate/autofix against the changed-file list** (the Epic→Story diff) instead of `.`. `format-autofix-scoped.js` already computes a changed-file set — the check gate could use the same list. This sidesteps cwd/ignore-glob interactions entirely.
2. **Run biome from the repo root** (main checkout) with an explicit path scope to the worktree's changed files, rather than `cwd=worktree` + `.`.
3. At minimum, **distinguish "0 files matched" from "formatting drift"** so an empty match is not reported as a gate failure — and never as a silent pass.

## Cross-references

- Consumer fix commit (athportal): `biome.json` anchor `!**/.worktrees` → `!.worktrees`.
- Prior consumer workaround that masked the bug (now reverted): added `--no-errors-on-unmatched` to `project.commands.formatCheck`/`formatWrite` in `.agentrc.json`.
- Surfaced during `/epic-deliver 12`; blocked Stories dsj1984/athportal#1116, dsj1984/athportal#1117, dsj1984/athportal#1118 at the format gate (all three implementations were correct).

dsj1984:
Closes #3407

### DIFF
--- a/.agents/scripts/lib/close-validation.js
+++ b/.agents/scripts/lib/close-validation.js
@@ -15,7 +15,7 @@
  * commit atomically with the Story PR.
  */
 
-import { spawn } from 'node:child_process';
+import { execFileSync, spawn } from 'node:child_process';
 import { writeFile as defaultWriteFile } from 'node:fs/promises';
 import { defaultGetHeadSha } from './close-validation/projections/head-sha.js';
 import { getCommands } from './config/commands.js';
@@ -33,6 +33,7 @@ import {
  * @property {string}   cmd   - Executable to run.
  * @property {string[]} args  - Arguments passed to `cmd`.
  * @property {string}   [hint] - Remediation hint shown on failure.
+ * @property {{ baseRef: string }} [changedFileScope] - Optional Story-diff scope.
  * @property {(cmd: string, args: string[], opts: { cwd: string, gateName?: string, log?: (m: string) => void, signal?: AbortSignal }) => Promise<{ status: number }> | { status: number }} [run]
  *   - Optional in-process runner. Story #1973: when present, the gate
  *     executes via this callable instead of spawning `cmd`/`args` through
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #3407

**Comparison**: `main` … `story-3407`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #3531, opened 2026-06-03
Guard postinstall + scope /.agents/ gitignore so the framework repo doesn't clobber/ignore its own .agents/ source (#3489)

### LINKED ISSUE #3489: Guard postinstall + scope /.agents/ gitignore so the framework repo doesn't clobber/ignore its own .agents/ source
## Context

Epic #3436 (npm distribution, PR #3485) added, in **this framework repo's own** root:

- `package.json` → `scripts.postinstall: "node bin/postinstall.js mandrel sync"` (Story #3469)
- `.gitignore` → `/.agents/` (Story #3469)

Both are correct for a **consumer** project, where `.agents/` is a regenerated working copy materialized from `node_modules/@mandrel/agents/`. But in the Mandrel source repo, `.agents/` **is** the committed product.

## Problem

1. **postinstall footgun:** running `npm install` here *without* `--ignore-scripts` would invoke `mandrel sync`, materializing `node_modules/@mandrel/agents/.agents/` over the committed source tree. Today this is masked only by the repo `.npmrc` (`ignore-scripts=true`); a contributor running `npm install --ignore-scripts=false` (or a tool that does) would clobber the source of truth.
2. **gitignore blind spot:** `/.agents/` in the framework repo's `.gitignore` silently ignores *new* `.agents/` files in future framework work. Already-tracked files are unaffected (so no current breakage), but a newly added `.agents/**` file could be missed in `git add`.

Surfaced as a 🟡 Medium in the Epic #3436 audit-results comment; escalated rather than fixed inline because it needs a dogfood-vs-ship design decision.

## Recommended fix

- Guard `bin/postinstall.js` to **no-op in the source checkout** — detect the Mandrel repo itself (e.g. `package.json#name === "@mandrel/agents"` resolved at the repo root, or a `.git` + repo-identity check) and exit 0 without materializing.
- Scope the `/.agents/` ignore to the **consumer-facing template / bootstrap output**, not the framework root — or add a negation (`!/.agents/` is wrong here; prefer not ignoring it in this repo at all and only emitting the ignore into consumer `.gitignore`).

## Acceptance (suggested)

- `npm install --ignore-scripts=false` in a clean Mandrel checkout does not modify tracked `.agents/` files.
- A newly created `.agents/**` source file is visible to `git status` in the framework repo.
- Consumer-side behavior (regenerate `.agents/` on install, ignore it in their VCS) is unchanged.

Refs Epic #3436, Stories #3469.

dsj1984:
Closes #3489

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,11 +5,14 @@ dist
 .DS_Store
 *.log
 
-# Regenerated agent payload (npm-distribution model): ./.agents/ is a working
-# copy materialized from node_modules/@mandrel/agents by `mandrel sync` (run by
-# the package postinstall hook). The package tarball is the source of truth, so
-# the regenerated copy is never committed. See Tech Spec #3459 / Feature #3461.
-/.agents/
+# NOTE: This is the Mandrel framework SOURCE repo, where ./.agents/ IS the
+# committed product — so it is deliberately NOT ignored here (Story #3489).
+# A blanket `/.agents/` ignore (correct for a *consumer* project, where
+# ./.agents/ is a working copy materialized from node_modules/@mandrel/agents
+# by `mandrel sync`) would silently hide newly added framework `.agents/**`
+# source files from `git status`/`git add`. Consumer projects add the
+# `/.agents/` ignore to their OWN .gitignore; the framework repo must not.
+# Local overrides under .agents/ are still ignored below.
 
 # Agent Local Overrides
 .agentrc.local.json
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #3489

**Comparison**: `main` … `story-3489`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #3577, opened 2026-06-04
fix(update): auto-detect the package manager in `mandrel update` — hardcoded `npm install` breaks & corrupts pnpm/yarn workspaces (#3575)

### LINKED ISSUE #3575: fix(update): auto-detect the package manager in `mandrel update` — hardcoded `npm install` breaks & corrupts pnpm/yarn workspaces
## Story

As an operator on a **pnpm (or yarn) workspace**, I want `npx mandrel update`
to upgrade `@mandrelai/agents` using the repo's actual package manager, so that
the one-shot `update` command works without manual workarounds and without
corrupting `node_modules`.

## Summary

`mandrel update` (bump → sync → migrate → doctor) is unusable in this repo
because its **install step hardcodes `npm install`** with no package-manager
detection. In a pnpm workspace `npm install` crashes, the bump never lands, and
— worse — the half-completed `npm install` leaves `node_modules` pointing at a
stale version until `pnpm install` repairs it.

> Note: the **original** Windows `spawnSync npm ENOENT` probe failure is already
> fixed in **v1.47.0** (`defaultVersionRunner` now passes
> `shell: process.platform === 'win32'`). This story is about the *remaining*
> install-step problem, not the probe.

## Environment

- OS: Windows 11, Node.js v24.11.1, pnpm 11.5.0
- Repo: pnpm workspace root (`pnpm-workspace.yaml` present, **no**
  `package-lock.json`); `@mandrelai/agents` in root `devDependencies`
- `@mandrelai/agents` versions observed: 1.45.0 → 1.47.0 → 1.48.0

## Reproduction

```
$ npx mandrel update
Updating v1.47.0 → v1.48.0…
npm error Cannot read properties of null (reading 'matches')
Error: mandrel update: install command `npm install @mandrelai/agents@1.48.0` exited 1:
    at defaultNpmUpdate (.../lib/cli/update.js:304:11)
```

`npm install` is the wrong tool in a pnpm workspace (there is no
`package-lock.json`; npm chokes on the pnpm-managed tree), so it exits 1 and
`defaultNpmUpdate` throws.

## Root cause (v1.47.0 `lib/cli/update.js`)

1. **Hardcoded `npm install` default** — `resolveInstallCmd(target, override)`
   (≈ line 263) returns `npm install ${PACKAGE_NAME}@${target}` whenever no
   `--install-cmd` override is supplied. There is no detection of
   `pnpm-lock.yaml` / `yarn.lock`. `defaultNpmUpdate` (≈ line 289) then runs
   that command and throws on the non-zero exit (line 304).

2. **`--install-cmd` can't carry the probed target** — the override is used
   *verbatim*, and the resolved `target` is appended **only when no override is
   supplied** (documented at lines 256–257, implemented at 263–265). So the
   escape hatch cannot be combined with the auto-probed "newest" version: an
   operator must hardcode the version
   (`--install-cmd "pnpm add -D @mandrelai/agents@1.48.0 -w"`), which defeats the
   purpose of `update`. There is no `{target}` placeholder.

3. **Non-atomic failure corrupts `node_modules`** — when the default
   `npm install` fails mid-flight in a pnpm workspace, it partially mutates the
   tree and flips the top-level `node_modules/@mandrelai/agents` symlink to a
   stale store entry (observed reverting 1.47.0 → 1.45.0). Subsequent `mandrel`
   invocations then execute the **wrong version** until `pnpm install` repairs
   the link. The failure should be atomic (no mutation) or self-healing.

## Acceptance Criteria

- [ ] **AC-1** In a pnpm workspace (`pnpm-lock.yaml` present), `npx mandrel
  update` with **no flags** detects pnpm and runs
  `pnpm add -D @mandrelai/agents@<target>`, adding `-w` when the manifest is the
  workspace root. The bump lands in `package.json` **and** `pnpm-lock.yaml` and
  the command exits 0.
- [ ] **AC-2** `yarn.lock` present → `yarn add -D @mandrelai/agents@<target>`;
  neither pnpm nor yarn lockfile → today's `npm install` behavior. The npm path
  is unchanged for npm repos.
- [ ] **AC-3** `--install-cmd` continues to work as an explicit override and
  supports a documented placeholder (e.g. `{target}`) so an override can still
  consume the auto-probed newest version; OR detection makes the override
  unnecessary for the common case and the docs say so.
- [ ] **AC-4** If the install step fails for any reason, `node_modules` is left
  in its prior consistent state (no half-mutation / no flipped symlink), or the
  thrown error tells the operator to run their package manager's install to
  repair. `mandrel update` must never silently leave a stale linked version.
- [ ] **AC-5** The existing v1.47.0 win32 `shell: true` handling on both the
  probe (`defaultVersionRunner`) and the install spawn is preserved.

## Verification

- [ ] On Windows + a pnpm workspace root, run `npx mandrel update` from an
  out-of-date pin; assert `package.json` + `pnpm-lock.yaml` advance to newest and
  exit code is 0, then `sync`/`migrate`/`doctor` run as normal.
- [ ] Re-run `npx mandrel update` when already latest; assert idempotent
  no-op / "up to date".
- [ ] Negative: force the install command to fail; assert
  `node_modules/@mandrelai/agents` still resolves to the pre-run version
  (no corruption).
- [ ] Regression: an npm-only repo (with `package-lock.json`) still upgrades via
  `npm install` exactly as before.

## Workaround (until fixed)

Bypass the broken bump step and drive the package manager directly, then run the
Windows-safe standalone subcommands:

```
pnpm add -D @mandrelai/agents@latest -w
npx mandrel sync
npx mandrel migrate --from <old> --to <new>
npx mandrel doctor
```

dsj1984:
Closes #3575

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,12 @@ npx mandrel update
    [`docs/upgrade-major.md`](docs/upgrade-major.md), and exits non-zero
    without touching anything. Re-run with `--major` to apply it.
 3. **No-op short-circuit** — already on the newest version ⇒ nothing to do.
-4. **Install** the target version (the dependency bump is left **staged**
-   on disk — `mandrel update` performs no `git add` / `git commit`, so you
-   review and commit the lockfile change yourself).
+4. **Install** the target version with the project's package manager —
+   auto-detected from the lockfile (`pnpm-lock.yaml` ⇒ pnpm, `yarn.lock` ⇒
+   yarn, otherwise npm) so the bump lands in your real lockfile. The
+   dependency bump is left **staged** on disk — `mandrel update` performs no
+   `git add` / `git commit`, so you review and commit the lockfile change
+   yourself.
 5. **Sync** — re-materialize `./.agents/` from the freshly installed payload.
 6. **Migrate** — apply version-keyed migration steps for the crossed range.
 7. **Doctor** — run the check registry to verify the resulting install.
@@ -102,12 +105,14 @@ npx mandrel update
   runs.
 - `--major` — apply a major-version crossing that the gate would otherwise
   refuse. Review [`docs/upgrade-major.md`](docs/upgrade-major.md) first.
-- `--install-cmd "<cmd>"` — override the install command for pnpm/yarn
-  workspaces. The default is `npm install @mandrelai/agents@<target>`; pass
-  e.g. `--install-cmd "pnpm add @mandrelai/agents@<target>"` so the bump
-  lands in your real lockfile rather than writing a stray
-  `package-lock.json`. The registry probe always stays on `npm view` (it is
-  a PM-agnostic registry query).
+- `--install-cmd "<cmd>"` — override the auto-detected install command. The
+  package manager is normally detected from your lockfile
+  (`pnpm-lock.yaml` ⇒ `pnpm add -D …`, `yarn.lock` ⇒ `yarn add -D …`,
+  otherwise `npm install …`), so an override is rarely needed. When you do
+  pass one, a `{target}` placeholder is substituted with the resolved newest
+  version — e.g. `--install-cmd "pnpm add -D @mandrelai/agents@{target} -w"` —
+  so the override can still consume the auto-probed version. The registry
+  probe always stays on `npm view` (it is a PM-agnostic registry query).
 
 ### Manual equivalent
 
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #3575

**Comparison**: `main` … `story-3575`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #3582, opened 2026-06-04
fix(postinstall): source-checkout guard misfires on every consumer install, skipping `.agents/` sync (#3580)

### LINKED ISSUE #3580: fix(postinstall): source-checkout guard misfires on every consumer install, skipping `.agents/` sync
## Story

As an operator installing `@mandrelai/agents` as a dependency, I want the
published `postinstall` hook to actually materialize `./.agents/` into my
project, so that a fresh install is usable without manually running
`mandrel sync`.

## Summary

The `isSourceCheckout` guard in [`bin/postinstall.js`](bin/postinstall.js)
is broken for **every consumer install**: it resolves the package's *own*
`package.json` instead of the consumer repo root, so its
`name === "@mandrelai/agents"` check always matches and the hook
short-circuits to `exit(0)` **before** any sync runs. The `.agents/` payload
is never materialized on install; consumers silently land in the
`mandrel doctor`-detected "not materialized" state.

## Root cause (`bin/postinstall.js`)

The source-checkout detector resolves the repo root from the module's own
location:

```js
// isSourceCheckout(), lines 71-75
const here = path.dirname(fileURLToPath(moduleUrl));      // .../node_modules/@mandrelai/agents/bin
const pkgPath = path.join(here, '..', 'package.json');    // .../node_modules/@mandrelai/agents/package.json
const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
return pkg?.name === PACKAGE_NAME;                         // ALWAYS @mandrelai/agents
```

When installed as a dependency the file lives at
`<consumer>/node_modules/@mandrelai/agents/bin/postinstall.js`, so `..`
resolves to the **package's own** `package.json` — whose `name` is, of course,
`@mandrelai/agents`. The guard fires on every consumer install and
`runPostinstall` exits 0 at line 113-116 before invoking `runSync`.

In the framework source checkout the same logic happens to work (`..` lands at
`<repo-root>/package.json`, also named `@mandrelai/agents`), which masked the
defect. The doc comment's claim that resolving "from this module's own
location ... so the detection is stable" is precisely the bug: the module's
own location is the *package*, never the consumer root.

## Proposed fix

Restore the original intent (compare the **consumer/repo root**
`package.json#name`) using two complementary, fail-safe signals:

1. **`node_modules` path guard (primary).** If `bin/postinstall.js` resolves
   to a path containing a `node_modules` segment, we are unambiguously
   installed as a dependency → **not** a source checkout → run sync. This is
   true by construction for every dependency install and needs no env var.

2. **`INIT_CWD` name check (fallback for the source repo).** npm sets
   `INIT_CWD` to the directory where `npm install` was invoked — the consumer
   root on a dependency install, or the framework repo root when installing
   the framework itself. Read `INIT_CWD/package.json` (falling back to the
   module-relative root only when `INIT_CWD` is unset) and treat it as a
   source checkout only when its `name === "@mandrelai/agents"`.

Both branches fail **safe** toward running sync (matching the file's
best-effort contract): any read/parse error or unset env returns `false`.
`INIT_CWD` must be an injectable seam (default arg) so the tests can drive
both branches.

## Scope of work

1. **Fix `isSourceCheckout`** in [`bin/postinstall.js`](bin/postinstall.js):
   add the `node_modules` path guard first, then the `INIT_CWD`-rooted
   package-name check, with `initCwd` as an injectable default-arg seam.
2. **Update the doc comment** (lines ~49-66 and ~19-30): correct the now-false
   "resolve from this module's own location" rationale to describe the
   consumer-root resolution, and document the `node_modules` + `INIT_CWD`
   signals.
3. **Add/adjust tests** in
   [`tests/cli/postinstall.test.js`](tests/cli/postinstall.test.js):
   - consumer install (module under `node_modules`) → `isSourceCheckout`
     returns `false` and `runSync` is invoked;
   - source checkout (`INIT_CWD` root named `@mandrelai/agents`, not under
     `node_modules`) → returns `true` and sync is skipped with `exit(0)`;
   - fail-safe: missing/malformed `package.json` and unset `INIT_CWD` →
     returns `false` (sync runs).

## Acceptance Criteria

- [ ] **AC-1** A consumer install (package resolved under
  `node_modules/@mandrelai/agents`) materializes `./.agents/` — `runSync`
  is invoked and the source-checkout guard does **not** fire.
- [ ] **AC-2** The framework source checkout (root `package.json#name` ===
  `@mandrelai/agents`, not under `node_modules`) still no-ops with `exit(0)`
  and never clobbers the committed `.agents/` source, including under
  `npm install --ignore-scripts=false`.
- [ ] **AC-3** The hook remains best-effort: any read/parse error, unset
  `INIT_CWD`, or sync failure degrades to `exit(0)` (with the existing hint on
  sync failure), never failing the consumer's `npm install`.
- [ ] **AC-4** The doc comment no longer claims module-relative resolution is
  stable for consumers; it accurately describes the `node_modules` + `INIT_CWD`
  signals.

## Verify

- [ ] `npm run test:quick` (or targeted
  `node --test tests/cli/postinstall.test.js`) passes, including the new
  consumer-install and source-checkout cases.
- [ ] Manual: in a scratch consumer project, `npm install <packed tarball>`
  materializes `./.agents/` (or run `npm pack` + install into a temp dir and
  confirm `.agents/` appears).
- [ ] `npm run lint` passes.

## Affected files

- `bin/postinstall.js`
- `tests/cli/postinstall.test.js`

chrisbarrantes:
Closes #3580


chrisbarrantes:
## 🔬 Code Review — Story #3580

**Comparison**: `main` … `story-3580`
**Provider**: `native`
**Findings**: 2

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 1
- 🟢 Suggestion: 1

### 🟡 Medium Risk (1)

#### 🟡 Size/Volume Warning — `.agents/scripts/lib/config-settings-schema.js` _[maintainability]_

Module `.agents/scripts/lib/config-settings-schema.js` reports a size/volume warning (module 49.4, worst method 49.4).

### 🟢 Suggestion (1)

PR #3585, opened 2026-06-04
fix(postinstall): sync writes .agents/ into node_modules, not consumer root (process.cwd() under npm lifecycle) (#3584)

### LINKED ISSUE #3584: fix(postinstall): sync writes .agents/ into node_modules, not consumer root (process.cwd() under npm lifecycle)
## Story

As an operator installing `@mandrelai/agents` as a dependency, I want the
`postinstall` hook to materialize `.agents/` into **my project root**, so that
a plain `npm install @mandrelai/agents` leaves a usable `./.agents/` without
my having to run `mandrel sync` by hand.

## Summary

The `postinstall` hook reaches `runSync`, but `runSync` derives its
destination from `process.cwd()` — and npm runs a dependency's lifecycle
scripts with cwd set to **the package's own directory**
(`node_modules/@mandrelai/agents/`). So `.agents/` is written into
`node_modules/@mandrelai/agents/.agents/` (a self-copy onto the source) and
the consumer's project root gets nothing.

This is the **second** postinstall bug, exposed once #3580 fixed the
`isSourceCheckout` guard: before #3580 the guard always fired and the sync
never ran, so this never had a chance to misfire. With the sync now running on
consumer installs, the wrong-destination bug surfaces (reproduced on a real
`npm install @mandrelai/agents` into `/…/man-install-test/`).

## Root cause (`lib/cli/sync.js`)

```js
// runSync(), lines 134-166
cwd = () => process.cwd(),            // default seam
...
const projectRoot = cwd();           // = node_modules/@mandrelai/agents during postinstall
const destRoot = path.join(projectRoot, '.agents');  // → wrong tree
```

`projectRoot` feeds both `resolvePackageRoot(projectRoot)` and `destRoot`.
Under the npm postinstall lifecycle, `process.cwd()` is the package directory,
not the consumer root, so the materializer copies the payload back onto itself
and never reaches the consumer project.

## Why CI's Install Matrix did not catch it

The Install Matrix gate runs an **explicit `mandrel sync`** after install
(pack → install → `mandrel sync` → assert materialization). That manual sync
runs with cwd = consumer root and materializes correctly, so the gate is green
even though the *postinstall* path is broken. The bug only affects consumers
who rely on the automatic postinstall materialization.

## Proposed fix

Use `INIT_CWD` (npm sets it to the directory where `npm install` was invoked —
the consumer root) as the destination root for the postinstall-driven sync,
falling back to `process.cwd()` when running outside npm. Scope the change to
the **postinstall caller** so the manual `mandrel sync` / `mandrel update`
paths (which already run from the project root) are unchanged:

```js
// bin/postinstall.js — runPostinstall, new initCwd seam:
sync({
  cwd: () => initCwd || process.cwd(),   // initCwd = process.env.INIT_CWD
  exit: (code) => { if (code !== 0) syncFailed = true; },
});
```

`runSync` already exposes a `cwd` injectable seam, so `lib/cli/sync.js` itself
needs no change — only the postinstall caller plus an `initCwd` seam on
`runPostinstall` for testability.

## Scope of work

1. **Add an `initCwd` seam to `runPostinstall`** in
   [`bin/postinstall.js`](bin/postinstall.js) (default `process.env.INIT_CWD`)
   and pass `cwd: () => initCwd || process.cwd()` into the `sync(...)` call.
2. **Update the doc comment** to explain the npm-lifecycle-cwd hazard and why
   the postinstall path must resolve the destination from `INIT_CWD`.
3. **Add/adjust tests** in
   [`tests/cli/postinstall.test.js`](tests/cli/postinstall.test.js): assert
   the `sync` seam receives a `cwd` resolving to `INIT_CWD` when set, and to
   `process.cwd()` when `INIT_CWD` is unset.

## Acceptance Criteria

- [ ] **AC-1** When `runPostinstall` runs the sync, the `runSync` call
  receives a `cwd` that resolves to `INIT_CWD` (the consumer project root)
  when `INIT_CWD` is set.
- [ ] **AC-2** When `INIT_CWD` is unset (the hook run outside npm), the sync
  `cwd` falls back to `process.cwd()`.
- [ ] **AC-3** The manual `mandrel sync` / `mandrel update` paths are
  unchanged — `runSync`'s default `cwd` (`process.cwd()`) is untouched in
  `lib/cli/sync.js`.
- [ ] **AC-4** The hook remains best-effort: any sync failure still degrades
  to `exit(0)` with the existing hint, never failing the install.

## Verify

- [ ] `node --test tests/cli/postinstall.test.js` passes, including the new
  `cwd`/`INIT_CWD` cases.
- [ ] Manual: `npm install <packed tarball>` into a scratch consumer project
  (whose cwd-during-postinstall is the package dir) materializes `.agents/`
  into the **consumer root**, not into `node_modules/@mandrelai/agents/`.
- [ ] `npm run lint` passes.

## Affected files

- `bin/postinstall.js`
- `tests/cli/postinstall.test.js`

chrisbarrantes:
Closes #3584

### DIFF
--- a/bin/postinstall.js
+++ b/bin/postinstall.js
@@ -119,9 +119,22 @@ export function isSourceCheckout({
  * and we also catch any thrown error (e.g. a mid-copy filesystem fault), so
  * the postinstall lifecycle can never fail the consumer's `npm install`.
  *
+ * **Destination root (Story #3584).** `runSync` derives its destination from
+ * `process.cwd()`, but npm runs a dependency's lifecycle scripts with cwd set
+ * to the **package's own directory** (`node_modules/@mandrelai/agents`), not
+ * the consumer project root. Left unset, the materializer would copy
+ * `.agents/` back onto the package's own payload and the consumer's project
+ * root would get nothing. We pass `runSync` a `cwd` resolving to `INIT_CWD` —
+ * the directory where `npm install` was invoked (the consumer root) — falling
+ * back to `process.cwd()` only when the hook runs outside npm (`INIT_CWD`
+ * unset). This is scoped to the postinstall path: the manual `mandrel sync` /
+ * `mandrel update` paths already run from the project root and keep
+ * `runSync`'s default `process.cwd()` behaviour.
+ *
  * @param {{
  *   sync?: typeof runSync,
  *   isSourceCheckout?: typeof isSourceCheckout,
+ *   initCwd?: string,
  *   writeErr?: (s: string) => void,
  *   exit?: (code: number) => void,
  * }} [opts]
@@ -131,6 +144,7 @@ export function isSourceCheckout({
 export function runPostinstall({
   sync = runSync,
   isSourceCheckout: detectSource = isSourceCheckout,
+  initCwd = process.env.INIT_CWD,
   writeErr = (s) => process.stderr.write(s),
   exit = (code) => process.exit(code),
 } = {}) {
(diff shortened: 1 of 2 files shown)


chrisbarrantes:
## 🔬 Code Review — Story #3584

**Comparison**: `main` … `story-3584`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #3617, opened 2026-06-05
fix(git-cleanup): branches phase strands merged local ref when worktree removal fails (Windows lock) (#3598)

### LINKED ISSUE #3598: fix(git-cleanup): branches phase strands merged local ref when worktree removal fails (Windows lock)
## Summary

`git-cleanup`'s branches phase strands an **already-merged local branch ref** whenever the candidate's worktree removal fails. On Windows this fires routinely: the per-worktree `node_modules` (and leftover test/coverage handles) keep `git worktree remove --force` from succeeding, so the local (and remote) ref deletion is skipped and the run exits non-zero with a spurious "1 failure(s)". The merged branch is pure garbage at that point — its remote ref is already deleted by the PR auto-merge — yet it lingers in the local checkout after every test-running Story delivery.

## Root cause (code-read)

The branch reap couples ref deletion to worktree removal:

- [`lib/orchestration/git-cleanup/phases/branches.js`](.agents/scripts/lib/orchestration/git-cleanup/phases/branches.js) — `executeCleanup` loops candidates with:
  ```js
  if (!reapWorktree({ cand, removeWorktreeFn, ... })) continue;   // <-- skips the rest
  if (!reapLocalRef({ cand, deleteLocalFn, ... })) continue;
  if (remote) reapRemoteRef({ ... });
  ```
  A `false` from `reapWorktree` `continue`s the loop, so **`reapLocalRef` never runs**.
- [`lib/orchestration/git-cleanup/phases/branches-reap.js`](.agents/scripts/lib/orchestration/git-cleanup/phases/branches-reap.js) — `reapWorktree` pushes a failure and returns `false` whenever `wtRes.ok` is false.
- [`lib/orchestration/git-cleanup/phases/git-probes-ff.js`](.agents/scripts/lib/orchestration/git-cleanup/phases/git-probes-ff.js) — `removeWorktree` returns `{ ok:false, dirty:true }` only when **both** `git worktree remove` and `git worktree remove --force` exit non-zero. (Note: `dirty` here means "plain remove failed, tried force" — not git-status-dirty.)

So a worktree the framework *cannot physically remove* (Windows lock on `node_modules`/test artifacts, or a stale admin entry after `single-story-close`'s `reapOnSuccess` already removed the dir) blocks deletion of a branch ref that is otherwise safe to delete.

Two gaps compound:

1. **Coupling.** Local/remote ref deletion should not be gated on worktree removal succeeding. The candidate was selected precisely because its PR is merged (`branches.js` plan resolves `prNumber` / merge state); a failure to delete the *directory* shouldn't strand the *ref*.
2. **No Windows-lock recovery in this path.** [`lib/worktree/lifecycle/reap.js`](.agents/scripts/lib/worktree/lifecycle/reap.js) already implements Stage-1 `fs.rm` retries, a `--force` fallback, and a `pending-cleanup.json` handoff for exactly this Windows-lock failure mode — but the git-cleanup branches-phase `removeWorktree` is a separate, simpler implementation that has none of it.

## Repro

On Windows, deliver any Story whose close-validation runs the test/coverage gates (so the per-worktree tree holds open handles), then run the Step-6 cleanup:

```
node .agents/scripts/git-cleanup.js --execute --remote --yes --fast-forward-main --branches --include story-<id>
```

Observed: `❌ worktree … (forced — was dirty)`, `❌ 1 failure(s) during cleanup`, exit 1, and `git branch --list story-<id>` still present. (Reproduced delivering athportal Story #1285 / PR #1286.)

## Proposed fix

- **Decouple ref reap from worktree reap.** In `executeCleanup`, when `reapWorktree` fails, record the worktree-removal failure but still attempt `reapLocalRef` (and `reapRemoteRef` in `--remote`) for a candidate whose merged state is already established. A stranded merged ref is worse than a leftover directory.
- **Align the remover with the hardened path.** Give the branches-phase `removeWorktree` the same Windows-lock recovery as `worktree/lifecycle/reap.js` (Stage-1 `fs.rm` retries + `pending-cleanup` handoff), or have it delegate to that primitive, so a locked directory degrades to a deferred sweep instead of aborting the candidate.
- **Classify lock-class failures as non-fatal** so the run doesn't exit 1 when the only problem is an OS file lock on an already-merged branch's worktree.

## Acceptance criteria

- [ ] A merged candidate whose worktree removal fails still has its local ref deleted (and remote ref in `--remote` mode); no merged `story-<id>` ref is stranded.
- [ ] A Windows-lock-class worktree-removal failure degrades to the existing recovery / `pending-cleanup` handoff rather than aborting the candidate's ref reap.
- [ ] `git-cleanup` does not exit non-zero solely because a locked directory for an already-merged branch could not be physically removed (it surfaces a warning + deferred-cleanup signal instead).
- [ ] Regression coverage for: (a) worktree-remove fails but branch is merged → local/remote ref still deleted; (b) the failure is recorded as a warning/deferred entry, not a hard failure.

## Out of scope

The `drain-pending-cleanup` skill (handles the WorktreeManager manifest path, not the stranded-ref coupling); non-Windows behavior (the coupling fix applies cross-platform, but the lock recovery is Windows-specific).

dsj1984:
Closes #3598

### DIFF
--- a/.agents/scripts/git-cleanup.js
+++ b/.agents/scripts/git-cleanup.js
@@ -69,6 +69,7 @@ import {
 import {
   buildJsonEnvelope,
   computeExitCode,
+  renderDeferredLine,
   renderDryRun,
   renderExecutionLine,
   renderExecutionSummary,
@@ -106,6 +107,7 @@ export {
   probeAllPrs,
   probeLatestPr,
   probeMergedPr,
+  renderDeferredLine,
   renderDryRun,
   renderExecutionLine,
   renderExecutionSummary,
--- a/.agents/scripts/lib/orchestration/git-cleanup/phases/branches-reap.js
+++ b/.agents/scripts/lib/orchestration/git-cleanup/phases/branches-reap.js
@@ -15,14 +15,44 @@
  *     drop tracking refs left behind by remote deletes.
  *
  * Every helper records its outcome on the supplied accumulator arrays
- * and pushes failures onto `failures`; return values are booleans
- * indicating whether to continue with the next reap step.
+ * and pushes hard failures onto `failures`.
+ *
+ * ## Ref-reap is decoupled from worktree-reap (Story #3598)
+ *
+ * The candidate set is, by construction, already-merged (the planner
+ * resolves each candidate's merged PR / `git merge-base` state). A
+ * failure to physically remove the worktree *directory* must NOT strand
+ * the merged branch *ref*: a leftover directory is recoverable garbage,
+ * a stranded merged ref clutters every subsequent run. So `reapWorktree`
+ * no longer gates the ref reap — `executeCleanup` always proceeds to
+ * `reapLocalRef` / `reapRemoteRef` after it, regardless of the worktree
+ * outcome. Lock-class removal failures (Windows handles on
+ * `node_modules` / test artifacts) are recorded as **deferred** entries
+ * and handed off to the `pending-cleanup` manifest rather than pushed
+ * onto `failures`, so an OS file lock on an already-merged branch's
+ * directory does not make `git-cleanup` exit non-zero.
  *
  * @module lib/orchestration/git-cleanup/phases/branches-reap
  */
 
 const TAG = '[git-cleanup]';
 
+/**
+ * Attempt to remove a candidate's worktree directory. Records the outcome
+ * on `worktrees`. A successful (or absent) removal is silent. A failed
+ * removal is split by class:
+ *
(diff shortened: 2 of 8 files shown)


dsj1984:
## 🔬 Code Review — Story #3598

**Comparison**: `main` … `story-3598`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #3677, opened 2026-06-05
MODULE_TYPELESS_PACKAGE_JSON warning when running bundled .agents/scripts in consumers (no ESM boundary in materialized tree) (#3589)

### LINKED ISSUE #3589: MODULE_TYPELESS_PACKAGE_JSON warning when running bundled .agents/scripts in consumers (no ESM boundary in materialized tree)
## Summary

Running any of the `@mandrelai/agents` bundled scripts in a **consumer repo**
(e.g. `node .agents/scripts/bootstrap.js`) emits a Node.js warning on every
invocation:

```
(node:XXXXX) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of
file:///.../.agents/scripts/bootstrap.js is not specified and it doesn't
parse as CommonJS. Reparsing as ES module because module syntax was detected.
This incurs a performance overhead.
```

The warning is **cosmetic** — scripts run correctly — but it clutters script
and CI output and is easily mistaken for an error during onboarding/install.

> Note: this does **not** reproduce in the framework's own dev repo, because
> the root [`package.json`](package.json) already declares `"type": "module"`.
> It only bites **consumers**, whose root `package.json` has no `"type"` field
> and who have no ESM boundary inside the materialized `.agents/` tree.

## Root cause

The package ships ES Module scripts. Verified in-tree:

- **595** `.js` files under `.agents/scripts/`
- **592** use `import` / `export`
- the remaining handful use the ESM-only `createRequire(import.meta.url)` pattern

When a consumer's root `package.json` has no `"type"` field, Node defaults to
CommonJS, fails to parse the ESM syntax, then re-parses as ESM — producing the
warning **and a double-parse on every process start**. There is currently **no
nested `package.json`** anywhere under `.agents/` to scope the module type
(verified: `find .agents -name package.json` returns nothing), so the consumer
root is the nearest manifest Node consults.

## Impact

- **Functional:** none — scripts run correctly.
- **Performance:** minor. Each affected file is parsed twice per process start.
- **UX / noise:** warning text clutters script output and CI logs, and can be
  mistaken for an error during onboarding/install — eroding first-run trust.

## Trade-offs of the obvious fix (`"type": "module"` in consumer root)

Adding `"type": "module"` to the consumer's root `package.json` silences the
warning, but it is **repo-wide**, not scoped to `@mandrelai/agents`:

- ✅ Cleanest one-liner; clears the warning for all 595 files at once.
- ⚠️ Changes the default module system for **every** plain `.js` file in the
  consuming repo, including future code the user adds. New `.js` files must
  then use `import`/`export` or break.
- ⚠️ Can conflict with downstream tooling that expects CommonJS `.js` (older
  config files, some test runners / build scripts), forcing `.cjs` renames.
- ⚠️ The framework imposing a project-wide ESM default is surprising for
  consumers who never opted into it.

## Potential issues to consider

- The scaffolder / installer may not own (or want to mutate) the consumer's
  root `package.json` `"type"` field.
- Consumers with existing CommonJS code would be broken by a blanket
  `"type": "module"`.
- Per-file `.mjs` renames only help the few directly-invoked entry points; the
  ~590 internal modules would still warn unless their directory has its own
  ESM boundary.

## Potential solutions (rough order of preference)

1. **Ship an ESM boundary inside the package's own directory (recommended).**
   Add a minimal `.agents/scripts/package.json` containing
   `{ "type": "module" }`. Node resolves module type from the nearest
   `package.json`, so this scopes ESM to the framework's own files without
   touching the consumer's root manifest or affecting their other code.
   Most self-contained fix.
2. **Scaffolder sets `"type": "module"` in the generated consumer
   `package.json` — but only when creating it fresh / when no conflicting
   CommonJS is present**, and document the implication. Avoid silently editing
   an existing consumer manifest.
3. **Rename directly-invoked entry scripts to `.mjs`** (bootstrap, sync,
   quality scripts) and update the script paths. Scoped, but does **not** clear
   warnings from the ~590 internal modules.
4. **Document the warning as a known cosmetic issue** with the recommended
   consumer-side fix, if no packaging change is made.

## Recommendation

**Option 1.** A `{ "type": "module" }` marker inside the package's own script
directory resolves the warning at the source, keeps the change scoped to
framework-owned files, and avoids imposing a module-system default on consuming
repos. A separate verification confirmed that adding `.agents/scripts/package.json`
with `{ "type": "module" }` eliminates the warning with **zero consumer-side
changes**.

### Distribution note (verified)

The published-package `files` allowlist already includes `.agents/`
(`"files": [".agents/","bin/","lib/"]`), so a `.agents/scripts/package.json`
marker would be carried in the npm payload and materialized into the consumer's
`.agents/scripts/` on `mandrel sync` — present on install and surviving
reinstalls. **Acceptance must verify** that:

- the marker is part of the published tarball (`npm pack` contents), and
- `mandrel sync` copies it into the consumer's `.agents/scripts/` (the sync
  allowlist / copier does not skip nested `package.json` files), and
- the marker does not perturb publishing of the root package (npm should not
  treat the nested manifest as a separate workspace/package).

## Acceptance

- Running `node .agents/scripts/bootstrap.js --help` (and other bundled
  scripts) in a fresh npm-installed consumer emits **no**
  `MODULE_TYPELESS_PACKAGE_JSON` warning.
- `.agents/scripts/package.json` containing `{ "type": "module" }` is present in
  the published tarball and materialized by `mandrel sync`.
- No regression in the framework's own dev repo (root already declares
  `"type": "module"`; the nested marker is redundant-but-harmless there).

## Verify

1. `npm pack` the package and confirm `.agents/scripts/package.json` is in the
   tarball.
2. Install into a fresh consumer repo **without** a root `"type"` field; run
   `mandrel sync`.
3. `node .agents/scripts/bootstrap.js --help` → exits clean, no warning.
4. `node .agents/scripts/sync.js` (or other bundled entry points) → no warning.

dsj1984:
Closes #3589

### DIFF
--- a/.agents/scripts/package.json
+++ b/.agents/scripts/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}


dsj1984:
## 🔬 Code Review — Story #3589

**Comparison**: `main` … `story-3589`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #3678, opened 2026-06-05
fix(story-close): cd-out guard false-negatives on symlinked paths (macOS) — realpath both sides before comparing (#3672)

### LINKED ISSUE #3672: fix(story-close): cd-out guard false-negatives on symlinked paths (macOS) — realpath both sides before comparing
# fix(story-close): cd-out guard false-negatives on symlinked paths (macOS) — realpath both sides before comparing

## Context

The story-close **cd-out guard**
([`cd-out-guard.js`](.agents/scripts/lib/orchestration/story-close/cd-out-guard.js))
refuses to close a Story while the operator's shell is still `cd`'d into the
per-story worktree being reaped. It is a safety check: closing from inside the
worktree makes `--cwd` semantics impossible to honour (git ops target the main
repo while the filesystem reap targets the directory the caller is sitting in;
on Windows this surfaces as `EBUSY … rmdir`).

The guard's verdict is a **string comparison of two resolved paths**. That
comparison gives a **false negative on any host where the working path contains
a symlinked component** — most notably macOS, where `/tmp` → `/private/tmp` and
`/var` → `/private/var`. On such hosts the guard silently returns `{ ok: true }`
and **does not fire**, defeating the safety check.

This surfaces today as a permanently-red test on every macOS delivery:
`tests/story-close-cd-out-guard.test.js` →
*"exits 1 with the remediation message when CWD is the worktree being reaped"*
fails locally on macOS but passes in Linux CI. The red test has been worked
around with `--skip-validation` during recent Story deliveries (#3664, #3665),
which is exactly the kind of normalized gate-skipping this Story removes.

## Root cause

[`checkCdOutGuard`](.agents/scripts/lib/orchestration/story-close/cd-out-guard.js)
compares the two paths like this:

```js
// cd-out-guard.js:40-42
const workCwd = path.resolve(mainCwd, worktreeRoot, `story-${storyId}`);
const cwd = path.resolve(currentCwd);            // currentCwd defaults to process.cwd()
if (cwd !== workCwd) return { ok: true };        // <-- false-negative branch
```

The two operands are canonicalized **differently**:

- **`workCwd`** is built from `mainCwd`, which is the **verbatim `--cwd`
  argument** threaded through `resolveCloseInputs`. `path.resolve` normalizes
  `.`/`..` and makes the path absolute but **does not resolve symlinks**. So on
  macOS it stays `/tmp/…/.worktrees/story-<id>`.
- **`cwd`** defaults to `process.cwd()`, which the OS returns **fully
  canonicalized** (symlinks resolved). On macOS that becomes
  `/private/tmp/…/.worktrees/story-<id>`.

`'/tmp/…' !== '/private/tmp/…'` → the guard takes the `{ ok: true }` branch and
never fires.

### Why it manifests as a test failure (and only on macOS)

The execution order in
[`close-inputs.js`](.agents/scripts/lib/orchestration/story-close/close-inputs.js)
is:

```js
// close-inputs.js (abridged)
const config   = resolveConfig({ cwd });                 // :98
const guard    = checkCdOutGuard({ ... });               // :101
if (!guard.ok) throw new Error(guard.message);           // :107  <-- guard fires here
const provider = injectedProvider || createProvider(config); // :109
```

The subprocess test
([`story-close-cd-out-guard.test.js:137`](tests/story-close-cd-out-guard.test.js#L137))
spawns `story-close.js --story 999 --cwd <tmp>` with the child's `cwd` set to
`<tmp>/.worktrees/story-999`, and asserts the output matches `/Refusing to
close/`.

- **Linux CI:** `os.tmpdir()` is `/tmp` (not a symlink) → `workCwd === cwd` →
  guard fires at line 107 → `"Refusing to close"` prints → **test passes**.
- **macOS local:** `os.tmpdir()` resolves under `/private/tmp` via `process.cwd()`
  but `mainCwd` stays `/tmp` → comparison fails → guard no-ops → execution falls
  through to `createProvider(config)` at line 109, which throws
  `[ProviderFactory] provider is required …` because the temp dir has no
  `.agentrc.json`. The subprocess still exits 1 (so the status assertion
  passes), but the **message is the provider error, not the guard message**, so
  `assert.match(output, /Refusing to close/)` fails.

Node reports this as `fail 2` because it counts the failing leaf subtest **and**
its parent `describe` block (`story-close cd-out guard (subprocess)`); there is
**one** underlying defect. The pure-helper test
*"rejects when cwd matches the worktree being reaped"* passes because it feeds
two already-matching literal paths (`/repo/.worktrees/story-746`) with no
symlink component — it never exercises the canonicalization mismatch.

## Impact

- **Latent safety bug (not just test noise).** In real use, a macOS operator
  whose repo lives under a symlinked path (e.g. `/tmp`, `/var`, or a
  symlinked home/work dir) gets **no protection** from the cd-out guard — it
  silently fails to refuse a close-from-inside-the-worktree, which is the exact
  footgun the guard exists to prevent.
- **Normalized gate-skipping.** The red test trains deliveries to pass
  `--skip-validation`, eroding the local close-validation gate and masking any
  *other* failure that lands in the same suite.
- **CI blind spot.** Because Linux CI is green, the defect is invisible to the
  required-check gate and only bites contributors on macOS.

## Suggested fix

Canonicalize **both** operands the same way before comparing — resolve symlinks
on each side (falling back to `path.resolve` when the path does not yet exist,
since `realpathSync` throws on missing paths):

```js
import fs from 'node:fs';

function canonical(p) {
  try {
    return fs.realpathSync(p);
  } catch {
    return path.resolve(p); // path may not exist yet; normalized comparison is still correct
  }
}

const workCwd = canonical(path.resolve(mainCwd, worktreeRoot, `story-${storyId}`));
const cwd = canonical(currentCwd);
if (cwd !== workCwd) return { ok: true };
```

Keep the function pure and unit-testable: the `realpath` seam can be injected
(default `fs.realpathSync`) so the existing pure-helper tests can drive both the
symlinked and non-symlinked path without touching the filesystem.

## Acceptance Criteria

- [ ] `checkCdOutGuard` canonicalizes both `currentCwd` and the derived
      `workCwd` (symlinks resolved) before the equality check, with a safe
      fallback for not-yet-existing paths.
- [ ] The guard fires correctly when `currentCwd` and `mainCwd` differ only by a
      symlinked prefix (e.g. `/tmp` vs `/private/tmp`).
- [ ] `tests/story-close-cd-out-guard.test.js` passes on macOS **and** Linux —
      specifically the subprocess case *"exits 1 with the remediation message
      when CWD is the worktree being reaped"* asserts the `Refusing to close`
      message on both platforms.
- [ ] A new unit case exercises the symlink false-negative directly (inject a
      `realpath` seam that maps `/tmp/x` → `/private/tmp/x`) so the regression is
      caught without depending on the host OS.
- [ ] No `--skip-validation` is needed to close a Story on macOS solely because
      of this test; the full close-validation suite is green on macOS.

## Verify

1. On macOS, on `main` before the fix: `node --test tests/story-close-cd-out-guard.test.js`
   → `fail 2` (the subprocess remediation-message case).
2. Apply the fix; re-run on macOS → all pass.
3. Confirm Linux CI stays green (the change is a no-op where no symlink prefix
   exists).
4. Sanity-check the pure helper still rejects a literal worktree-match and still
   returns `ok` for single-tree mode / main-repo / sibling-worktree inputs.

## Out of Scope

- The downstream `[ProviderFactory] provider is required` error is a *symptom*
  of the guard not firing in this test's temp dir; it is not itself a defect to
  fix here.
- `git-cleanup` worktree-reap tests were investigated and **pass** on current
  `main` (likely already addressed by the ref-reap/worktree-removal decoupling
  in commit `819ebbcf`); no change is needed there. If a `git-cleanup` path is
  later found to do the same verbatim-`--cwd` vs `process.cwd()` comparison,
  file a separate follow-up.

## Notes

- Root cause confirmed by code-read + local repro on macOS (Darwin):
  `os.tmpdir()` → `/tmp/claude-501`, `fs.realpathSync` → `/private/tmp/claude-501`.
- Guard call site / ordering:
  [`close-inputs.js:98-109`](.agents/scripts/lib/orchestration/story-close/close-inputs.js#L98-L109).
- `acceptance::n-a` — this repo ships no `.feature` files.

dsj1984:
Closes #3672

### DIFF
--- a/.agents/scripts/lib/orchestration/story-close/cd-out-guard.js
+++ b/.agents/scripts/lib/orchestration/story-close/cd-out-guard.js
@@ -14,11 +14,41 @@
  * `workCwd` to the main repo, so the equality check is a tautology there
  * and we don't reject those.
  *
+ * Both operands are canonicalized through `fs.realpathSync` before the
+ * equality check (Story #3672). `path.resolve` normalizes `.`/`..` and makes
+ * a path absolute but does NOT resolve symlinks, while `process.cwd()` is
+ * returned fully canonicalized by the OS. On hosts where the working path
+ * contains a symlinked component — most notably macOS, where `/tmp` →
+ * `/private/tmp` and `/var` → `/private/var` — the two strings would never
+ * match, so the guard silently no-opped and failed to fire. Realpath'ing
+ * both sides closes that false-negative.
+ *
  * Pure: takes inputs, returns a verdict. Exported so the rejection path is
- * unit-testable without spawning the script.
+ * unit-testable without spawning the script. The `realpath` seam is injected
+ * (default `fs.realpathSync`) so tests can drive the symlinked and
+ * non-symlinked cases without touching the filesystem.
  */
 
+import fs from 'node:fs';
 import path from 'node:path';
+
+/**
+ * Canonicalize a path: resolve symlinks via `realpath`, falling back to
+ * `path.resolve` when the path does not exist yet (`realpathSync` throws on
+ * missing paths). The fallback keeps the comparison correct for
+ * not-yet-created worktree directories.
+ *
+ * @param {string} p
+ * @param {(p: string) => string} realpath
+ * @returns {string}
+ */
+function canonical(p, realpath) {
+  try {
+    return realpath(p);
+  } catch {
+    return path.resolve(p);
+  }
+}
 
 /**
  * @param {object} opts
@@ -27,6 +57,7 @@ import path from 'node:path';
  * @param {number|string} opts.storyId
  * @param {string} [opts.worktreeRoot]     `delivery.worktreeIsolation.root` (defaults to `.worktrees`).
  * @param {string} [opts.currentCwd]       Defaults to `process.cwd()`.
+ * @param {(p: string) => string} [opts.realpath]  Symlink-resolution seam (defaults to `fs.realpathSync`).
  * @returns {{ ok: true } | { ok: false, message: string }}
  */
 export function checkCdOutGuard({
@@ -35,10 +66,14 @@ export function checkCdOutGuard({
   storyId,
   worktreeRoot = '.worktrees',
   currentCwd = process.cwd(),
+  realpath = fs.realpathSync,
 }) {
   if (!cwdExplicit) return { ok: true };
-  const workCwd = path.resolve(mainCwd, worktreeRoot, `story-${storyId}`);
-  const cwd = path.resolve(currentCwd);
+  const workCwd = canonical(
+    path.resolve(mainCwd, worktreeRoot, `story-${storyId}`),
+    realpath,
+  );
+  const cwd = canonical(currentCwd, realpath);
   if (cwd !== workCwd) return { ok: true };
   return {
     ok: false,
--- a/tests/story-close-cd-out-guard.test.js
+++ b/tests/story-close-cd-out-guard.test.js
@@ -131,6 +131,80 @@ test('checkCdOutGuard pure helper', async (t) => {
     assert.equal(result.ok, false);
     assert.match(result.message, /story-42/);
   });
+
+  await t.test(
+    'fires when currentCwd and mainCwd differ only by a symlinked prefix (Story #3672)',
+    () => {
+      // Regression for the macOS false-negative: the derived `workCwd`
+      // (built from the verbatim `--cwd` arg via path.resolve, symlinks NOT
+      // resolved) and `process.cwd()` (OS-canonicalized) point at the same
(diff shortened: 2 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #3672

**Comparison**: `main` … `story-3672`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #3929, opened 2026-06-10
refactor(plan): re-point or remove vacated validators and dead init guards (#3906)

### LINKED ISSUE #3906: refactor(plan): re-point or remove vacated validators and dead init guards
## Context
Several planning/init validators are vacated or dead, giving false confidence (see `docs/epic-lifecycle-review.md` §1.6).

- `lib/orchestration/task-body-validator.js` skips any Story whose body is a string; the decompose-author skill MANDATES string bodies, so the verify-tier suffix rule, vague-verb check, and non-empty-goal check never run on canonical decompositions.
- `lib/story-init/dependency-guard.js` (~300 LOC) can never block under 3-tier (`tasks.length === 0 → continue`) and passes `config` where `orchestration` is destructured (so `repoSlug` is always undefined — proof it never ran).
- The dispatch focus-overlap engine (`dependency-analyzer.js`) always produces zero edges since Task deletion; the concurrency-gate has "no default loader wired" so it trivially passes.

## Acceptance
- `task-body-validator` either re-points at the parsed string body (via `parseStoryBody`) so its rules fire, or is deleted and the skill's "validator rejects" claims removed.
- The dead dependency-guard stage and focus-overlap engine are deleted (or wired to a real input); no validator advertises a guarantee it does not provide.

## Verify
- Unit tests for any retained validator prove the rules fire on string bodies.
- `npm test` green after deletions.

dsj1984:
Closes #3906


dsj1984:
## 🔬 Code Review — Story #3906

**Comparison**: `main` … `story-3906`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4036, opened 2026-06-11
fix(cli): mandrel update changelog surface — ship or fetch docs/CHANGELOG.md (not in tarball) (#4035)

### LINKED ISSUE #4035: fix(cli): mandrel update changelog surface — ship or fetch docs/CHANGELOG.md (not in tarball)
# Make `mandrel update`'s changelog surface work — ship or fetch the changelog

## Context

The final step of `mandrel update` is "surface changelog" — print the
`docs/CHANGELOG.md` section(s) for the applied range `(current, target]` so
the operator can reconcile consumer-side guidance against what changed. In a
real consumer upgrade v1.58.0 → v1.59.0 (`dsj1984/domio`) it printed:

```
mandrel update: changelog not found at
  …/mandrel@1.58.0_…/node_modules/mandrel/docs/CHANGELOG.md — skipping changelog surface.
```

There are **two** independent bugs here:

1. **Wrong-version path** — it resolves `docs/CHANGELOG.md` against the
   **old** version's install dir. This is the same stale-process root cause
   addressed by the sibling Story (re-exec post-install phases from the new
   bin); fixing that re-points the path at the new version.

2. **The changelog isn't in the published tarball** — even resolved against
   the correct version, the lookup fails because `docs/CHANGELOG.md` is **not
   shipped in the npm package**. Verified: the installed
   `mandrel@1.59.0` package dir contains no `docs/CHANGELOG.md` (only
   `.agents/rules/changelog-style.md`). The changelog exists only in the
   GitHub repo (`dsj1984/mandrel/docs/CHANGELOG.md`), which is where the
   operator had to fetch it manually via `gh api` to complete the
   `/agents-update` reconciliation step. **This bug is not fixed by the
   re-exec Story** — the file simply isn't there to read.

The workflow doc (`/agents-update`) tells operators "you do not need to fetch
a CHANGELOG from anywhere, since the CLI emitted it inline" — which is
currently false for npm consumers.

## Proposed change

Pick one (or both) so the surface step has a real source:

- **Ship it in the tarball.** Add `docs/CHANGELOG.md` to the package's
  `files` allowlist so it is published and resolvable at the installed
  package root. Simplest fix; makes the inline surface work offline.
- **Or fetch it at update time.** The update already needs network for the
  registry probe (`npm view`); fetch the changelog (or just the
  `(current, target]` range) from GitHub Releases / the raw repo when the
  packaged file is absent. More robust if shipping the full generated
  changelog in the tarball is undesirable.

Whichever path: resolve against the **target** version (depends on / pairs
with the re-exec Story), and degrade gracefully with an actionable message
(link to the GitHub releases page) when the changelog genuinely can't be
obtained — never a bare "not found … skipping."

## Acceptance

- After `mandrel update` v(N) → v(N+1), the surfaced changelog for the
  applied range is printed inline from a real source (packaged file or
  fetched), with no manual `gh api` / web lookup needed.
- When the changelog cannot be obtained, the message is actionable (points to
  the releases URL), not a silent skip.
- If shipping in the tarball: `npm pack` (or the publish dry-run) includes
  `docs/CHANGELOG.md`, asserted by a packaging test.

## Verify

- Unit test for the surface step's source resolution (packaged-present,
  packaged-absent→fetch, both-absent→actionable message).
- Manual: `mandrel update` across a minor bump prints the new version's
  changelog section inline.

## Notes

Reported from a live `/agents-update` run on `dsj1984/domio`. Sibling Story
covers the stale-process root cause (re-exec post-install phases from the new
bin), which fixes the wrong-version-path half; this Story covers the
packaging/availability half, which is independent.

dsj1984:
Closes #4035

### DIFF
--- a/lib/cli/__tests__/update-changelog-surface.test.js
+++ b/lib/cli/__tests__/update-changelog-surface.test.js
@@ -1,0 +1,357 @@
+// lib/cli/__tests__/update-changelog-surface.test.js
+/**
+ * Unit tests for the `defaultSurfaceChangelog` fallback chain and
+ * `fetchChangelogFromGitHub` introduced by Story #4035 — mandrel update
+ * changelog surface (ship or fetch).
+ *
+ * The three paths under test (via the `run` entrypoint):
+ *   1. Packaged file present → reads and prints the matching section(s).
+ *   2. Packaged file absent, GitHub fetch succeeds → prints the section(s)
+ *      from the fetched content.
+ *   3. Both sources unavailable → emits an actionable warning with a link
+ *      to the GitHub Releases page; never throws.
+ *
+ * Plus unit coverage for `fetchChangelogFromGitHub` itself:
+ *   - Resolves the content from the first tag that returns 2xx.
+ *   - Tries the namespaced tag (`mandrel-v<ver>`) first, then bare (`v<ver>`).
+ *   - Throws when both tag forms return non-2xx.
+ *   - Throws when the HTTP request errors out.
+ *
+ * `defaultSurfaceChangelog` is exercised through the `run` default export
+ * so the full wiring from `deps` through to output is verified. The
+ * `fetchChangelog` seam (and `https` in `fetchChangelogFromGitHub`) are
+ * injected so no real HTTP call occurs (testing-standards § Unit: mock all I/O).
+ *
+ * Tier: unit (testing-standards § Unit). All I/O — filesystem and network
+ * — is mocked via injectable seams.
+ *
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4035

**Comparison**: `main` … `story-4035`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4037, opened 2026-06-11
fix(cli): mandrel update materializes stale .agents/ — re-exec post-install phases from the new bin (#4034)

### LINKED ISSUE #4034: fix(cli): mandrel update materializes stale .agents/ — re-exec post-install phases from the new bin
# Re-exec `mandrel update` post-install phases from the newly-installed binary

## Context

`mandrel update` runs its full step plan — `npm-update → runSync →
runMigrations → doctor → surface changelog` — **in a single process, as the
currently-installed (old) binary**. After the `npm-update` step swaps
`node_modules/mandrel` to the target version, the still-running old process
continues the remaining phases using the **already-loaded old module** (Node
cannot hot-swap a `require`d module mid-process). So `runSync`,
`runMigrations`, `doctor`, and the changelog step all execute the *old*
code and resolve paths against the *old* package install dir.

Observed during a real consumer upgrade v1.58.0 → v1.59.0 (`dsj1984/domio`):

- **Stale `.agents/` materialization (silent, serious).** `runSync`
  materialized the **1.58.0** payload even though the on-disk package was
  already 1.59.0. The consumer ended up with `package.json` / lockfile pinned
  to 1.59.0 but `.agents/` carrying 1.58.0 content. Concretely, the
  in-process sync reported "Materialized **821** files"; re-running
  `mandrel sync` from the freshly-installed 1.59.0 bin materialized **832** —
  11 files the update had missed, including the entire new
  `/audit-documentation` workflow and `check-arch-cycles.js`.
- **`doctor` false-green.** The in-process `doctor` reported a clean
  `agents-drift` (old payload checked against itself), masking the stale
  tree. A consumer who trusts the "✅ Updated / doctor passed" output ships a
  version-mismatched `.agents/` and never knows.
- **Changelog path resolved to the old version.** The surface step looked
  for `docs/CHANGELOG.md` under `…/mandrel@1.58.0_…/` (the old install dir)
  and skipped (see the sibling Story for the packaging half of that bug).

The root cause is the in-process model: every phase after `npm-update` must
run against the **newly-installed** package, not the loaded old module.

## Proposed change

After the `npm-update` step lands the new version, re-exec the post-install
phases from the newly-installed binary rather than continuing in the stale
process:

- `spawn` / re-exec `mandrel sync`, `mandrel migrate` (if a standalone
  surface exists; otherwise the relevant phase entrypoint), and
  `mandrel doctor` from the just-installed `node_modules/.bin/mandrel` (or
  re-resolve the package root with a cache-busted `require`/dynamic import so
  the payload, migration set, and doctor checks come from the new version).
- Keep the orchestration/UX in the parent process; only the version-sensitive
  phases need to run from the new bin.
- Ensure the final `doctor` (especially `agents-drift`) runs from the new bin
  so a stale sync can never pass as green again.

This single change fixes the stale-`.agents/` materialization, the
`doctor` false-green, and the old-version path resolution that also breaks
the changelog surface.

## Acceptance

- After `mandrel update` v(N) → v(N+1) on a consumer, `.agents/` is
  materialized from the **v(N+1)** payload (file count + content match the
  installed package), with no second manual `mandrel sync` required.
- `mandrel doctor`'s `agents-drift` check run **by the update itself**
  reflects the newly-installed payload (a stale sync fails the in-update
  doctor instead of passing).
- `runMigrations` applies the target version's migration set, not the old
  process's.
- A regression test simulates the install-then-continue boundary (e.g. a
  fixture where the new payload adds a file) and asserts the post-install
  phases observe the new payload.

## Verify

- Unit/integration test around the update orchestration (`lib/cli/update.js`
  or equivalent) covering the re-exec boundary.
- Manual: upgrade a consumer across a version that adds a `.agents/workflows/*`
  file; confirm the new workflow + its `.claude/commands/` entry appear with
  no manual re-sync, and the in-update `doctor` is green only because the tree
  is genuinely correct.

## Notes

Reported from a live `/agents-update` run. Pairs with the changelog-packaging
Story (the changelog surface has a second, independent bug: the changelog
is not shipped in the published npm tarball at all).

dsj1984:
Closes #4034

### DIFF
--- a/lib/cli/__tests__/update-reexec.test.js
+++ b/lib/cli/__tests__/update-reexec.test.js
@@ -1,0 +1,513 @@
+// lib/cli/__tests__/update-reexec.test.js
+/**
+ * Unit tests for the Story #4034 re-exec fix in lib/cli/update.js.
+ *
+ * After `npm-update` lands the new version, the post-install phases
+ * (sync, migrate, doctor) MUST run from the newly-installed binary rather
+ * than in the still-running old process. The `spawnPhase` seam is the
+ * injectable boundary that makes this fully testable without a real npm
+ * install.
+ *
+ * Coverage contract (Story #4034):
+ *   - When `spawnPhase` is injected, post-install phases run through it
+ *     (re-exec path) instead of the in-process runSync/runMigrations/runDoctor.
+ *   - Phases are spawned in the correct order: sync → migrate → doctor.
+ *   - `mandrel migrate` receives `--from <current>` and `--to <target>`.
+ *   - The new bin path passed to `spawnPhase` is resolved from the cwd's
+ *     `node_modules/.bin/mandrel` (the production path).
+ *   - A failing sync phase throws and exits non-zero (never silently continues).
+ *   - A failing migrate phase throws and exits non-zero.
+ *   - A non-zero doctor phase maps to `action: 'doctor-failed'` + exit 1.
+ *   - `resolveNewBinPath` returns the correct platform path.
+ *   - `defaultSpawnPhase` routes stdout/stderr through the write sinks and
+ *     throws on spawn error.
+ *
+ * Tier: unit (testing-standards § Unit). All seams — including `spawnPhase`
+ * — are injected; no real child process, filesystem, or network call occurs.
+ *
+ * Security (security-baseline § 5 — Data Leakage & Logging): fixtures carry
+ * only version strings and paths; no tokens, credentials, or env values.
+ */
+
+import assert from 'node:assert/strict';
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4034

**Comparison**: `main` … `story-4034`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4066, opened 2026-06-12
fix: `mandrel update` no-op short-circuit leaves a stale `.agents/` unhealed (gates on installed version, not materialized drift) (#4065)

### LINKED ISSUE #4065: fix: `mandrel update` no-op short-circuit leaves a stale `.agents/` unhealed (gates on installed version, not materialized drift)
# fix: `mandrel update` no-op short-circuit leaves a stale `.agents/` unhealed (gates on installed version, not materialized drift)

## Context

`mandrel update` is documented (README § Update) as the one command that
advances the package **and** re-materializes `./.agents/`. But its no-op
short-circuit gates purely on the **installed package version**, so it
cannot heal a consumer whose package is already at the newest version while
`./.agents/` is materialized from an older payload.

`lib/cli/update.js` (`runUpdate`, lines ~813–825):

```js
const target = String(await resolveTargetVersion());     // newest published
// current = defaultCurrentVersion() → node_modules/mandrel/package.json version
if (compareVersions(target, current) <= 0) {
  write(`✅  Already up to date (v${current} is the newest version).\n`);
  return { ok: true, action: 'up-to-date', stepsRun: [], ... };  // returns BEFORE runSync
}
```

When `installedVersion === newestPublished`, the function returns before
`runSync` / `sync-commands` / `migrate` / `doctor` run. The check conflates
"package installed at newest" with "consumer materialized at newest" — they
are not the same whenever the dependency bump landed via any path other than
`mandrel update` itself.

### How a consumer reaches this state (reproduced in a real repo)

1. A renovate-style `chore: update mandrel` PR bumps `package.json` +
   lockfile only (no re-materialize). Two such bumps (v1.59.0 → v1.61.0)
   landed this way.
2. The package's best-effort `postinstall` hook (`bin/postinstall.js`, which
   runs `mandrel sync`) **did not fire**: the consumer uses **pnpm 10**, which
   blocks dependency build/lifecycle scripts by default unless the package is
   listed in `pnpm.onlyBuiltDependencies`. `mandrel` was not allowlisted, so
   `.agents/` was never auto-materialized. (This is the common case for pnpm-v10
   consumers, not an edge case.)
3. `npx mandrel update` (and `/agents-update`, which wraps it) then printed
   `✅ Already up to date (v1.61.0 is the newest version)` and exited 0 —
   never re-materializing. `.agents/` stayed at the v1.59.0 payload
   (pre-`/plan`+`/deliver` consolidation, 3-tier hierarchy, stale `SDLC.md` /
   `full-agentrc.json`) while `mandrel doctor` correctly reported
   `agents-drift`.
4. Only an explicit `npx mandrel sync` (the step the short-circuit skips)
   fixed it: 835 files materialized, 76 stale pruned, doctor → 10/10.

This is the **same failure class** Story #4034 fixed for the upgrade path
("silent stale-`.agents/` materialization and doctor false-green observed in
the v1.58.0 → v1.59.0 consumer upgrade"). That fix re-execs `runSync` from
the new binary — but **only when versions differ**. The no-op path has the
identical failure mode and is uncovered.

## Proposed Fix

Make the no-op short-circuit drift-aware rather than version-only. When
`installedVersion >= newestPublished`, do not unconditionally return — first
check whether `./.agents/` is actually materialized to the installed payload
(reuse the same signal `doctor`'s `agents-drift` check uses), and if drift is
detected, run `runSync` + `sync-commands` (and surface what changed) even
though the version is unchanged. Only emit `Already up to date` when both the
version is current **and** `.agents/` matches the installed payload.

Equivalent acceptable shape: keep the version short-circuit but have it run a
`doctor` `agents-drift` probe and fall through to a sync-only heal path on
drift, so `mandrel update` becomes idempotently self-healing regardless of how
the dependency version arrived.

Secondary (separate, smaller follow-up — not required by this Story): have
`mandrel init` / `mandrel sync` docs and the consumer bootstrap surface the
pnpm-v10 `pnpm.onlyBuiltDependencies` requirement (or have the doctor
`agents-drift` remedy mention it), so the postinstall hook actually runs on
pnpm-v10 consumers and the auto-materialize path is not silently disabled.

## Acceptance Criteria

- [ ] `mandrel update` re-materializes `./.agents/` (runSync + sync-commands)
      when the installed package version already equals the newest published
      version **but** `./.agents/` is drifted from the installed payload —
      i.e. the no-op short-circuit no longer returns before the sync phases
      when drift is present.
- [ ] When the version is current **and** `./.agents/` matches the installed
      payload, `mandrel update` still prints `Already up to date` and makes no
      changes (true no-op preserved; idempotent).
- [ ] After `mandrel update` completes on a drifted-but-current consumer,
      `mandrel doctor` reports `agents-drift` passing (10/10) with no manual
      `mandrel sync` required.
- [ ] The drift detection reuses the existing `agents-drift` doctor signal /
      payload-comparison helper rather than introducing a second drift
      definition.
- [ ] `--dry-run` on a drifted-but-current consumer reports that a sync/heal
      will run (not `Already up to date`), and writes nothing.

## Verification

- [ ] New unit test in `lib/cli/__tests__/update*.test.js`: with
      `currentVersion === target` and an injected drift signal, `runUpdate`
      invokes the sync phase(s) and returns `action: 'updated'` (or a new
      `'resynced'` action) instead of `'up-to-date'`.
- [ ] New unit test: version current **and** no drift → still
      `action: 'up-to-date'`, `stepsRun: []`.
- [ ] Manual repro: on a consumer pinned to the newest version with a
      hand-staled `./.agents/`, `npx mandrel update` heals it to 10/10 doctor
      in one command, with no separate `mandrel sync`.

---

_Filed from the domio consumer repo after hitting this in practice. Concrete
artifacts: doctor flagged `agents-drift` while `mandrel update --dry-run`
reported `Already up to date (v1.61.0)`; `npx mandrel sync` materialized 835
files / pruned 76 and cleared the drift._

dsj1984:
Closes #4065

### DIFF
--- a/lib/cli/__tests__/update-drift.test.js
+++ b/lib/cli/__tests__/update-drift.test.js
@@ -1,0 +1,360 @@
+// lib/cli/__tests__/update-drift.test.js
+/**
+ * Unit tests for Story #4065: drift-aware no-op short-circuit in
+ * lib/cli/update.js.
+ *
+ * The no-op short-circuit previously gated only on the installed package
+ * version. When `installedVersion === newestPublished` it returned immediately,
+ * leaving a drifted `.agents/` unhealed. This file verifies the new behaviour:
+ *
+ *   1. Version current + drift detected  → runUpdate invokes sync phase(s) and
+ *      returns `action: 'resynced'` (or `action: 'dry-run'` under --dry-run).
+ *   2. Version current + no drift        → runUpdate still returns
+ *      `action: 'up-to-date'` with `stepsRun: []` (true no-op preserved).
+ *   3. --dry-run + drift                 → reports that a heal will run;
+ *      emits no "Already up to date" message; writes nothing.
+ *
+ * Tier: unit (testing-standards § Unit). All seams including `checkDrift` are
+ * injected; no real filesystem, package, or network call occurs.
+ *
+ * Security (security-baseline § 5): fixtures contain only version strings and
+ * boolean drift signals; no credentials, tokens, or env values.
+ */
+
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+
+import { runUpdate } from '../update.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/** Capture stdout/stderr writes and exit code. */
+function makeCapture() {
+  const out = [];
+  const err = [];
+  let exitCode = null;
+  return {
+    out,
+    err,
+    get exitCode() {
+      return exitCode;
+    },
+    write: (s) => out.push(s),
+    writeErr: (s) => err.push(s),
+    exit: (code) => {
+      exitCode = code;
+    },
+  };
+}
+
+/**
+ * Minimal seam set for a "version already current" scenario. The injected
+ * `checkDrift` toggle lets us probe both the drifted and non-drifted branches
+ * without touching the real filesystem.
+ *
+ * @param {{ drifted?: boolean }} [opts]
+ */
+function makeUpToDateSeams({ drifted = false } = {}) {
+  const calls = [];
+  return {
+    calls,
+    currentVersion: '1.61.0',
+    resolveTargetVersion: async () => {
+      calls.push('resolveTargetVersion');
+      return '1.61.0'; // same as current — no-op path
+    },
+    checkDrift: async () => {
+      calls.push(`checkDrift:${drifted}`);
+      return drifted;
+    },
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4065

**Comparison**: `main` … `story-4065`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4094, opened 2026-06-13
perf: serial blocked-by edge creation during Epic decomposition (#4082)

### LINKED ISSUE #4082: Serial blocked-by edge creation during Epic decomposition
## Goal
Serial blocked-by edge creation during Epic decomposition

1. [MEDIUM] (latency) **Serial blocked-by edge creation during Epic decomposition** — providers/github/blocked-by-add.js` → `applyBlockedByDependencies() (landed in the top commit, refs #4067) walks every Story with a non-empty `dependsOn` array fully serially. For each Story it issues one `GET …/dependencies/blocked_by` round-trip (`fetchExistingBlockedBy`) and then a serial `for (blockerId of …)` loop of `POST` round-trips (`addBlockedByEdges`, lines 92–110). The Stories themselves are also processed one-at-a-time in the outer `for (story of stories)` loop (line 155). For an Epic with `M` Stories carrying dependency edges and `E` total edges, this is `M` serial GETs + `E` serial POSTs back-to-back — every request blocks the next, so total wall-clock latency is the *sum* of all round-trip times. Blocker `getTicket(…)` lookups are cache-mitigated (the provider's per-instance cache means a blocker referenced by many Stories is only fetched once), so this is serial-I/O latency, **not** an unbounded N+1. The function runs once at decompose time, but for a large Epic that is the difference between a sub-second and a multi-second tail on the planning flow. The remediation primitive already exists in-tree: `lib/util/concurrent-map.js` (`concurrentMap`, bounded, order-preserving, first-rejection-wins) — and is already used a few files over in `providers/github/sub-issues.js`.

## Acceptance
- [ ] Serial blocked-by edge creation during Epic decomposition — Replace the two serial loops with `concurrentMap` at a modest cap (e.g. 4–6) to respect GitHub's secondary rate limits while collapsing the latency from `sum(round-trips)` toward `sum(round-trips) / concurrency`. Run the per-Story work (the GET + its POST batch) under one `concurrentMap` over `stories`, and the per-edge POSTs under a nested `concurrentMap` over `blockerInternalIds`. The function's existing per-edge / per-Story `try/catch` + summary-counter contract (never throws, counts `added/skipped/failed`) is preserved because `concurrentMap` surfaces a single first error and the bodies already swallow their own. Expected gain: roughly linear in the chosen concurrency for any Epic with more than a handful of dependency edges.

## Agent Prompts

**Serial blocked-by edge creation during Epic decomposition**

```
In .agents/scripts/providers/github/blocked-by-add.js, parallelize applyBlockedByDependencies and addBlockedByEdges using the existing bounded primitive from lib/util/concurrent-map.js (import { concurrentMap }). In addBlockedByEdges, replace the serial 'for (const blockerId of …)' POST loop with a concurrentMap over the missing (non-skipped) blocker ids at concurrency 4–6, aggregating added/failed counts from the per-item results; keep the existing per-edge try/catch warn-and-count behavior so the function still never throws. In applyBlockedByDependencies, replace the outer 'for (const story of stories)' loop with a concurrentMap over stories (same cap), summing edgesAdded/edgesSkipped/edgesFailed/storiesProcessed from each result. Preserve the existing Logger.warn messages, the idempotent fetch-existing-first behavior, and the no-op-on-empty short-circuits. Add or extend a unit test that asserts the aggregated counts are identical to the current serial implementation for a fixture with multiple stories sharing a common blocker, and that getTicket is still memoized (no extra fetches for a repeated blocker).
```

## Context

This Story was opened by `/audit-to-stories` from the following audit reports:

- [`temp/audits/audit-performance-results.md`](temp/audits/audit-performance-results.md)

dsj1984:
Closes #4082

### DIFF
--- a/.agents/scripts/providers/github/blocked-by-add.js
+++ b/.agents/scripts/providers/github/blocked-by-add.js
@@ -21,7 +21,15 @@
  */
 
 import { Logger } from '../../lib/Logger.js';
+import { concurrentMap } from '../../lib/util/concurrent-map.js';
 import { parseApiJson } from './request-helpers.js';
+
+/**
+ * Bounded concurrency for the GitHub dependency-edge round-trips. Kept modest
+ * to respect GitHub's secondary rate limits while still collapsing the wall-
+ * clock latency from `sum(round-trips)` toward `sum(round-trips) / concurrency`.
+ */
+const EDGE_CONCURRENCY = 5;
 
 /**
  * Fetch the existing blocked-by issue numbers for a given issue.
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4082

**Comparison**: `main` … `story-4082`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4103, opened 2026-06-13
fix: 2-tier Stories un-initializable: composeStoryBody omits `Epic: #N` when epicId === parentId, blocking the /deliver wave loop (#4102)

### LINKED ISSUE #4102: 2-tier Stories un-initializable: composeStoryBody omits `Epic: #N` when epicId === parentId, blocking the /deliver wave loop
## Summary

Under the 2-tier ticket hierarchy (Epic → Story, Stories attached **directly** to the Epic), `composeStoryBody` never writes the `Epic: #N` footer line, but `story-init.js` requires that exact line to resolve the Story's Epic. The result: **every directly-attached 2-tier Story is un-initializable** — `story-init.js` aborts, blocking the entire `/deliver` wave loop.

## Root cause

Two pieces of the Story-body contract have drifted out of sync since the 2-tier collapse (Epic #4041 / decisions.md "Adopt a 2-tier hierarchy").

**1. The body producer suppresses the `Epic:` line when `epicId === parentId`.**

`.agents/scripts/providers/github/tickets.js` — `composeStoryBody` (line ~72):

```js
const lines = ['---', `parent: #${parentId}`];
if (epicId !== undefined && epicId !== null && epicId !== parentId) {
  lines.push(`Epic: #${epicId}`);
}
```

And `TicketGateway.createTicket` (line ~313) coerces a missing epicId to the parent:

```js
const epicId = ticketData.epicId || parentId;
```

In the 3-tier era a Story's `parent` was a **Feature** (distinct from the Epic), so `epicId !== parentId` held and the `Epic:` line was emitted. Under 2-tier, Stories attach **directly** to the Epic, so `parentId === epicId`, the guard is false, and the `Epic: #N` line is **never written**. The body gets only `parent: #N`.

**2. The body consumer resolves the Epic *only* from a literal `Epic:` line — no fallback to `parent:`.**

`.agents/scripts/lib/story-lifecycle.js` — `resolveStoryHierarchy` (line ~29):

```js
const epicMatch = source.match(/(?:^\s*epic:\s*#(\d+))/im);
const parentMatch = source.match(/(?:^\s*parent:\s*#(\d+))/im);
return {
  epicId: epicMatch ? Number.parseInt(epicMatch[1], 10) : null,
  parentId: parentMatch ? Number.parseInt(parentMatch[1], 10) : null,
};
```

It does **not** fall back to `parent:` when no `Epic:` line is present.

The abort is in `.agents/scripts/lib/story-init/context-resolver.js` (line ~46):

```js
const { epicId, parentId } = resolveStoryHierarchy(body);
if (!epicId) {
  throw new Error(
    `Story #${storyId} has no "Epic: #N" reference in its body. Cannot resolve hierarchy.`,
  );
}
```

So for any 2-tier Story, `composeStoryBody` writes `parent: #N` only → `resolveStoryHierarchy` returns `epicId: null` → `story-init.js` throws → the `/deliver` wave loop cannot dispatch the Story.

## Consumer impact

athportal (mandrel **v1.62.0**) Epic #23: all 5 decomposed Stories (#1529–#1533) were created with `parent: #23` and **no** `Epic: #23` line. `/deliver` could not initialize any of them. Worked around by **hand-patching all 5 Story bodies** to add an `Epic: #23` line.

## Intended contract

`docs/architecture.md` (~line 739) is explicit:

> Stories without an `Epic: #N` reference do not enter the Epic wave loop.

i.e. the Epic-attached wave loop is **gated on the presence of an `Epic: #N` reference**, and (per decisions.md) under 2-tier the Story's `parent` *is* the Epic. The `Epic: #N` ref should therefore always be present on an Epic-attached Story. The `epicId !== parentId` guard is a **stale 3-tier assumption** (Story-parent-is-a-Feature) that no longer holds.

## Candidate fixes

**A. (Recommended) Remove the `epicId !== parentId` guard in `composeStoryBody`** so the `Epic: #N` line is always emitted whenever an `epicId` is known:

```js
const lines = ['---', `parent: #${parentId}`];
if (epicId !== undefined && epicId !== null) {
  lines.push(`Epic: #${epicId}`);
}
```

Rationale: keeps `Epic: #N` as the single authoritative seam the rest of the runtime already reads (`resolveStoryHierarchy`, the wave loop's architecture-documented gate, `story.heartbeat`/`acceptance-eval` `epicId` fields). It restores the producer to match the documented contract instead of teaching every consumer a new fallback. Minimal blast radius: one conditional, one producer.

**B. Make `resolveStoryHierarchy` fall back to `parent:` when no `Epic:` line is present** (treat `parent` as the Epic ref under 2-tier):

```js
epicId: epicMatch ? Number.parseInt(epicMatch[1], 10)
        : (parentMatch ? Number.parseInt(parentMatch[1], 10) : null),
```

Rationale: also valid and arguably more robust (it heals already-created Stories without a body rewrite), but it spreads the "parent doubles as Epic" assumption into the resolver and any other reader, and it diverges from the architecture doc's "`Epic: #N` reference" wording. If chosen, the `Standalone multi-Story delivery (no Epic)` path must be re-checked: a standalone Story legitimately has a `parent` but no Epic, and must continue to resolve `epicId: null` so it routes to `helpers/deliver-stories.md` rather than the Epic wave loop — so a blind `parent:` fallback would **mis-route standalone Stories into the Epic loop**. This is the key reason to prefer Fix A.

**Recommendation: Fix A.** It is the smallest change, keeps a single source of truth (`Epic: #N`), matches the architecture doc verbatim, and does not risk mis-routing standalone Stories.

## Existing tests encode the buggy behavior

`tests/providers/github/tickets.test.js` ("composeStoryBody — single blocked-by footer (Story #3958)") currently **asserts the absence** of the `Epic:` line when `parentId === epicId`:

```js
const out = composeStoryBody({ ..., parentId: 1477, epicId: 1477, ... });
assert.match(out, /---\nparent: #1477$/);              // no Epic: line
```

with the comment "Epic line present because epicId !== parentId". These assertions must be updated under Fix A to expect `---\nparent: #1477\nEpic: #1477`.

`tests/lib/story-lifecycle.test.js` has **no** case for a body containing only `parent:` and no `Epic:` line — that gap is exactly what let the regression through.

## Suggested regression test

- `composeStoryBody` unit test: when `epicId === parentId`, the rendered body **includes** `Epic: #N` (the directly-attached 2-tier case). Update the Story #3958 assertions accordingly.
- Round-trip test: `resolveStoryHierarchy(composeStoryBody({ parentId: 23, epicId: 23 }))` returns `{ epicId: 23, parentId: 23 }` (non-null `epicId`). This nails the producer↔consumer contract that drifted.
- `context-resolver` test: a Story body produced for a directly-attached 2-tier Story does **not** throw `has no "Epic: #N" reference`.

dsj1984:
Closes #4102


dsj1984:
## 🔬 Code Review — Story #4102

**Comparison**: `main` … `story-4102`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4204, opened 2026-06-16
perf: mandrel doctor payload-drift check reads every materialized file twice (no stat short-circuit) (#4193)

### LINKED ISSUE #4193: mandrel doctor payload-drift check reads every materialized file twice (no stat short-circuit)
## Goal
2. `mandrel doctor` payload-drift check fully reads every materialized file twice (no stat short-circuit)

1. [MEDIUM] (efficiency (i/o)) **2. `mandrel doctor` payload-drift check fully reads every materialized file twice (no stat short-circuit)** — `lib/cli/registry.js` (the `agents-materialized` / payload-drift check, loop at `registry.js:529`) compares the installed package `.agents/` payload against the consumer's materialized `.agents/` tree by reading **both** the source bytes and the destination bytes of **every** payload file and comparing with `srcBytes.equals(destBytes)` (`registry.js:544-547`). On the common "clean" path (no drift) it reads the **entire payload twice into memory** — hundreds of files — before returning "match". It short-circuits only on the *first* difference, so a clean tree pays the full 2N-read cost. This runs on every `mandrel doctor` invocation and on the **Install Matrix CI gate** (`install-matrix.yml`), which asserts a `doctor` ready verdict on every release-PR leg.

## Acceptance
- [ ] 2. `mandrel doctor` payload-drift check fully reads every materialized file twice (no stat short-circuit) — Gate the byte read on a cheap `statSync` size comparison first: if `statSync(src).size !== statSync(dest).size`, the files are definitionally different — report drift without reading either file's contents. Only when sizes match do you fall through to the `readFileSync` + `.equals()` byte comparison (sizes can match while content differs, so the byte check is still required for the equal-size case). For a clean tree, size is equal on every file, so this does not save the clean path — but it makes the **drifted** path (the case the operator most wants fast feedback on, and the case where a size-changed file is overwhelmingly the common drift shape) return on stat alone. To additionally speed the clean path, compare a streamed hash or use `fs.read` of the first/last block before full read; or accept the clean-path cost since it is bounded and infrequent. **Expected magnitude:** for the drifted case, replaces up to 2 full-file reads-to-completion with 2 `stat` syscalls when the drift is a size change (the typical case after a `sync`). Low risk, additive guard.

## Agent Prompts

**2. `mandrel doctor` payload-drift check fully reads every materialized file twice (no stat short-circuit)**

```
In lib/cli/registry.js, the payload-drift check (the function containing the loop `for (const rel of files)` around line 529 that compares sourceRoot vs destRoot) reads full file bytes of both src and dest for every payload file via fsImpl.readFileSync and compares with srcBytes.equals(destBytes). Add a stat-size short-circuit before the byte read: stat both src and dest (via fsImpl.statSync, threading statSync through the injected fsImpl seam — extend the fsImpl typedef and the default to include statSync), and when the sizes differ, report drift immediately (same { ok: false, detail, remedy } shape) WITHOUT reading file contents. Only when the sizes are equal, fall through to the existing readFileSync + .equals() byte comparison (equal size does not guarantee equal content). Preserve the existing short-circuit-on-first-difference behavior, the comparedCount/missingCount bookkeeping, and all returned detail/remedy strings. Add a unit test that injects an fsImpl whose statSync reports a size mismatch and asserts the byte readFileSync is NOT called for that file. ---
```

## Context

This Story was opened by `/audit-to-stories` from the following audit reports:

- [`temp/audits/audit-performance-results.md`](temp/audits/audit-performance-results.md)

dsj1984:
Closes #4193

### DIFF
--- a/lib/cli/registry.js
+++ b/lib/cli/registry.js
@@ -465,6 +465,39 @@ function runAgentsMaterialized({ cwd, existsSync, resolvePackage } = {}) {
 // (Story #4048 B3 — no mirror copies).
 
 /**
+ * Decide whether a single materialized payload file has drifted from its
+ * package-payload source.
+ *
+ * Size short-circuit (Story #4193): when the injected `fsImpl` exposes
+ * `statSync`, a `statSync(src).size !== statSync(dest).size` mismatch is
+ * sufficient proof of drift, so it returns `true` on two `stat` syscalls
+ * **without reading either file's contents** — the overwhelmingly common
+ * drift shape after a stale `sync` is a size change. Equal sizes do not
+ * guarantee equal content, so an equal-size pair falls through to the
+ * `readFileSync` + `Buffer.equals` byte comparison. The short-circuit is
+ * opt-in on `statSync` being present; a seam that omits it (e.g. a legacy
+ * test double) transparently uses the byte-read path, preserving the prior
+ * behaviour exactly.
+ *
+ * Security: returns only a boolean — it never surfaces file contents to the
+ * caller (security baseline §5 — Data Leakage & Logging).
+ *
+ * @param {string} src - absolute path to the package-payload file
+ * @param {string} dest - absolute path to the materialized file
+ * @param {{ readFileSync: Function, statSync?: Function }} fsImpl
+ * @returns {boolean} `true` when the files differ
+ */
+function payloadFileDrifted(src, dest, fsImpl) {
+  if (
+    typeof fsImpl.statSync === 'function' &&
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4193

**Comparison**: `main` … `story-4193`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4230, opened 2026-06-17
feat: story-plan tech-stack hydrator is hard-bound to architecture.md §Tech Stack and misses numbered/relocated headings (#4228)

### LINKED ISSUE #4228: story-plan tech-stack hydrator is hard-bound to architecture.md §Tech Stack and misses numbered/relocated headings
## Context

`readTechStackSummary` in [`.agents/scripts/lib/story-plan.js`](.agents/scripts/lib/story-plan.js) populates the `techStack` field of the standalone-Story planning envelope. It is hard-bound to a single source and a strict heading shape:

```js
const archPath = path.join(projectRoot, 'docs', 'architecture.md');
// ...
const match = content.match(/^##\s+Tech Stack\s*$([\s\S]*?)(?=^##\s+\S)/m);
return match ? `## Tech Stack${match[1]}`.trim() : null;
```

Two real consumer-project observations show this is too brittle:

1. **Numbered headings silently yield `null`.** A consumer whose architecture.md uses `## 1. Tech Stack` (a perfectly ordinary numbered-section convention) never matches `^##\s+Tech Stack\s*$`, so `techStack` is `null` and the planner loses the stack summary — with no warning. This is a live case in the `athportal` consumer today.

2. **Relocating the inventory to its own doc isn't supported.** A consumer that splits the stack inventory into `docs/tech-stack.md` (single-ownership: WHAT in tech-stack.md, HOW in architecture.md, WHY in the ADRs) has no way to point the hydrator at the new file. The `sibling swarm-os` consumer did exactly this split; it only keeps working because it left a `## Tech Stack` pointer-stub heading in architecture.md, so the hydrator returns the stub text ("See tech-stack.md…") rather than the real table. The dedicated `tech-stack.md` registered in `docsContextFiles` is what actually carries the inventory into context — the envelope field is now a redirect, not the inventory.

The doc references that name the source location are also hard-bound to `docs/architecture.md`:
- [`.agents/workflows/helpers/plan-story.md`](.agents/workflows/helpers/plan-story.md) (`techStack` row: "The `## Tech Stack` section of `docs/architecture.md`")
- [`.agents/instructions.md`](.agents/instructions.md) §1.G
- [`.agents/personas/devops-engineer.md`](.agents/personas/devops-engineer.md)

## Proposal

Teach the hydrator to resolve the stack inventory robustly, and update the three doc references to match (one hard cutover, per the contract-cutover policy — no shape-toleration branch left behind):

1. **Prefer a dedicated `docs/tech-stack.md`** when present (the emerging single-ownership convention), falling back to the `## Tech Stack` section of `docs/architecture.md` when it is not.
2. **Tolerate numbered/decorated headings** — match `## 1. Tech Stack`, `## Tech Stack`, etc. (e.g. relax the anchor to allow an optional leading `N.` / numbering token before `Tech Stack`).
3. **Harden the section-terminator** so a Tech Stack section that is the last `##` in the file still matches (the current lookahead `(?=^##\s+\S)` requires a following heading).
4. **Update the three doc references** and `plan-story.md`'s envelope-field table to describe the new resolution order.
5. Consider sourcing the path from `project.docsContextFiles` / a config key rather than a literal, so consumers don't fork the resolver.

## Acceptance Criteria

- [ ] `readTechStackSummary` returns the inventory from `docs/tech-stack.md` when that file exists, else from the architecture.md Tech Stack section.
- [ ] A numbered heading (`## 1. Tech Stack`) resolves correctly instead of returning `null`.
- [ ] A Tech Stack section that is the final `##` in the file resolves correctly.
- [ ] `plan-story.md` (envelope-field table), `instructions.md` §1.G, and `devops-engineer.md` are updated to the new resolution order in the same change (hard cutover — no dual-source tolerance left behind).
- [ ] Unit coverage for: dedicated-file present, numbered heading, last-section heading, and neither-present (`null`) cases.

## Notes

Surfaced while splitting `tech-stack.md` out in the `athportal` consumer (its `.agents/` tree is synced and intentionally not hand-edited, so the consumer-side workaround is to keep a pointer-stub heading + register `tech-stack.md` in `docsContextFiles`). This issue tracks the upstream fix so consumers don't need that workaround. Observed against `mandrel` 1.70.0 (athportal) and 1.72.0 (swarm-os).

dsj1984:
Closes #4228

### DIFF
--- a/.agents/instructions.md
+++ b/.agents/instructions.md
@@ -86,11 +86,13 @@ apply by checking the `.agents/rules/` directory (e.g.,
 ### G. Structured Configuration
 
 Refer to `.agentrc.json` to understand your operational limits (e.g., allowed
-auto-run permissions, default personas). Refer to the **Tech Stack** section
-of `docs/architecture.md` for the project's specific technology choices
-(database, ORM, API framework, auth provider, validation library, workspace
-paths). Project-specific technology context is intentionally kept out of
-`.agentrc.json`.
+auto-run permissions, default personas). For the project's specific
+technology choices (database, ORM, API framework, auth provider, validation
+library, workspace paths), refer to the project's Tech Stack inventory: a
+dedicated `docs/tech-stack.md` when present (the single-ownership convention),
+otherwise the **Tech Stack** section of `docs/architecture.md` (a numbered or
+decorated heading such as `## 1. Tech Stack` is fine). Project-specific
+technology context is intentionally kept out of `.agentrc.json`.
 
 ### H. Observability & Friction Telemetry
 
--- a/.agents/personas/devops-engineer.md
+++ b/.agents/personas/devops-engineer.md
@@ -14,8 +14,10 @@ express it through code or documented configuration files.
 
 1. **Read Context:** Before making any infrastructure changes, analyze the
    existing CI/CD, deployment, and security configurations.
-2. **Follow Protocols:** Adhere strictly to the **Tech Stack** section of
-   `docs/architecture.md` and the `orchestration` block of `.agentrc.json`.
+2. **Follow Protocols:** Adhere strictly to the project's Tech Stack
+   inventory — a dedicated `docs/tech-stack.md` when present, otherwise the
+   **Tech Stack** section of `docs/architecture.md` — and the `orchestration`
+   block of `.agentrc.json`.
 3. **Validate Always:** For every task, determine how the change will be
    monitored and validated (logs, health checks, or test gates).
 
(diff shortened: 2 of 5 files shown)


dsj1984:
## 🔬 Code Review — Story #4228

**Comparison**: `main` … `story-4228`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4244, opened 2026-06-19
feat: project .agents/local/workflows/ into slash commands (prune-exempt consumer commands) (#4243)

### LINKED ISSUE #4243: Project .agents/local/workflows/ into slash commands (prune-exempt consumer commands)
## Summary

Teach the command projector (`.agents/scripts/sync-claude-commands.js`) to treat
`.agents/local/workflows/` as a **second, prune-exempt projection source**, so a
consumer can author a durable, repo-committed slash command that survives
`npm install`, `mandrel sync`, and a full `mandrel update`.

**Motivating consumer:** `mandrel-bench` needs a durable `/benchmark` command. It
currently lives at `.agents/local/workflows/benchmark.md` (the only location that
survives) but is **not** projected, so it can't be invoked as `/benchmark`.

## Problem

Today a consumer cannot author a durable repo-local slash command — every
candidate location is reaped by Mandrel's own self-healing:

1. **`.agents/workflows/<name>.md`** → deleted by `mandrel sync`'s **sync-prune**
   (#4046): any file in the managed `.agents/` zone with no counterpart in the
   published payload is "stale" and removed on the next `sync` / `update`.
2. **`.claude/commands/<name>.md`** (hand-placed) → reaped by the projector,
   which `unlinkSync`s any command not backed by `.agents/workflows/`. It runs on
   **every `npm install`** (the `prepare` hook) and during `mandrel update`'s
   `sync-commands` phase. The tree is also gitignored — generated, never
   committed.

The one zone that **survives** is `.agents/local/` — the sanctioned consumer
additions space (#3498, `LOCAL_ZONE_DIR = 'local'` in `lib/cli/sync.js`), which
`mandrel sync` never copies into nor prunes. But the projector does **not** read
it, so material there never becomes a slash command.

### Why a consumer copy-script doesn't close the gap

A consumer can chain its own projector after Mandrel's in `package.json`
(`prepare` + `sync:commands`), and it self-heals on a plain `npm install`. But
`mandrel update` runs its phases in this order (`lib/cli/update.js`, #4046 A1c):

```
npm-update → sync → sync-commands → migrate → doctor
```

`sync-commands` (step 3) invokes the projector **directly from the new binary**,
not through the consumer's `package.json` scripts — so it runs **last** among the
command-affecting steps and reaps the consumer-copied command with no hook after
it to restore it. Net: a manual `npm run sync:commands` is required after every
update. The projector is the adversarial last writer; the fix belongs in it.

### File: `.agents/scripts/sync-claude-commands.js`

Today a single `SRC_DIR` (`.agents/workflows/`) feeds both the copy pass and the
prune pass:

```js
const SRC_DIR = process.env.SYNC_CLAUDE_COMMANDS_SRC
  ?? path.join(PROJECT_ROOT, '.agents', 'workflows');
// ...
const sources = fs.readdirSync(SRC_DIR, { withFileTypes: true })
  .filter(isTopLevelWorkflow).map((e) => e.name);
const sourceSet = new Set(sources);
for (const file of existing) {
  if (!sourceSet.has(file)) fs.unlinkSync(path.join(DEST_DIR, file)); // reap orphans
}
```

Proposed: enumerate an **ordered list** of source dirs and union their names into
`sourceSet`. Because `sourceSet` drives both projection **and** the orphan-reap, a
single change makes local commands both appear and survive:

```js
const PAYLOAD_SRC = process.env.SYNC_CLAUDE_COMMANDS_SRC
  ?? path.join(PROJECT_ROOT, '.agents', 'workflows');
const LOCAL_SRC   = path.join(PROJECT_ROOT, '.agents', 'local', 'workflows');
const SRC_DIRS = [PAYLOAD_SRC, LOCAL_SRC].filter(dirExists);

// enumerate each dir → [{ dir, name }]; payload first so it wins on collision
const entries = SRC_DIRS.flatMap((dir) =>
  fs.readdirSync(dir, { withFileTypes: true })
    .filter(isTopLevelWorkflow).map((e) => ({ dir, name: e.name })));

// collision policy: payload wins, warn on a shadowed local file
const byName = new Map();
for (const e of entries) {
  if (byName.has(e.name)) { Logger.warn(`  shadowed  ${e.name} (local copy ignored; payload wins)`); continue; }
  byName.set(e.name, e);
}
const sourceSet = new Set(byName.keys());   // ← prune-exempts local-projected commands too
```

The copy pass reads each file from its own `dir`. Optionally stamp a distinct
header for local-origin commands (e.g. `<!-- AUTO-GENERATED from .agents/local/ -->`).

### Design decisions

- **Collision policy: payload wins, warn.** A consumer must not silently shadow a
  core command (`/deliver`, `/plan`, …). Reverse only if "consumer override" is
  the desired semantic — that's a bigger call; default to safe.
- **Subdir name: `.agents/local/workflows/`** mirrors the payload's
  `.agents/workflows/` (rule becomes "top-level `.md` in any `workflows/` source
  dir"). `.agents/local/commands/` is an equally valid name if a
  `workflows`/`commands` split is preferred — pick one and document it.
- **No new prune risk.** `sync.js` already exempts the entire `.agents/local/`
  subtree; this change only affects the projector, which never writes into
  `.agents/local/` (local sources are read-only inputs).

## Acceptance criteria

1. A file at `.agents/local/workflows/foo.md` projects to `.claude/commands/foo.md`
   (invocable as `/foo`).
2. It **survives** `npm install`, `mandrel sync`, and a full `mandrel update`
   (`npm-update → sync → sync-commands → migrate → doctor`) with **no** manual
   re-sync.
3. A core payload command of the same basename **wins**; the local copy is ignored
   with a `shadowed` warning.
4. Removing the local file removes its projected command on the next sync (still
   reaped — no longer in any source).
5. The `SYNC_CLAUDE_COMMANDS_SRC` test override still works (existing fixture
   tests).

## Docs to touch

- `lib/cli/sync.js` / projector header comments — note the second source.
- `.agents/instructions.md` § 1.E (local overrides) — mention that
  `.agents/local/workflows/` now yields slash commands.
- Changelog entry under the consumer-extensibility theme.

---

_Filed from the `mandrel-bench` consumer, which hit this wall authoring a durable
`/benchmark` command. Until this ships, that command lives un-projected at
`.agents/local/workflows/benchmark.md` and is invoked by reference._

dsj1984:
Closes #4243

### DIFF
--- a/.agents/instructions.md
+++ b/.agents/instructions.md
@@ -76,6 +76,15 @@ environment variables that override project defaults. The config resolver
 deep-merges `.agentrc.local.json` over `.agentrc.json` (local wins; absent
 local file is a no-op). Do not modify these local files unless requested.
 
+**Durable slash commands.** Any `.md` file placed at
+`.agents/local/workflows/<name>.md` is automatically projected into
+`.claude/commands/<name>.md` by `sync-claude-commands.js`, making it
+invocable as `/<name>`. Because the entire `.agents/local/` subtree is
+exempt from `mandrel sync`'s prune pass, these commands survive
+`npm install`, `mandrel sync`, and `mandrel update` with no manual
+re-sync. Core payload commands of the same basename always win (the
+local copy is ignored with a `shadowed` warning).
+
 ### F. Modular Global Rules
 
 Before writing code or documentation, verify if any domain-agnostic rules
--- a/.agents/scripts/sync-claude-commands.js
+++ b/.agents/scripts/sync-claude-commands.js
@@ -1,10 +1,18 @@
 #!/usr/bin/env node
 
 /**
- * Projects .agents/workflows/ into a flat `.claude/commands/` tree so Claude
- * Code exposes each workflow as a bare `/<name>` slash command. The workflows
- * directory remains the single source of truth; this script is the only writer
- * of `.claude/commands/`.
+ * Projects .agents/workflows/ (and .agents/local/workflows/ when present) into
+ * a flat `.claude/commands/` tree so Claude Code exposes each workflow as a
+ * bare `/<name>` slash command. Two source directories are enumerated in order:
+ *
+ *   1. PAYLOAD_SRC  — `.agents/workflows/`  (the installed Mandrel payload)
+ *   2. LOCAL_SRC    — `.agents/local/workflows/`  (consumer-authored, prune-exempt)
+ *
+ * The payload directory wins on basename collision: if both sources supply
+ * `foo.md`, the payload copy is projected and the local copy is ignored with a
+ * `shadowed` warning. Because both sources are unioned into `sourceSet`, local
+ * commands are also protected from the orphan-reap — they survive
+ * `npm install`, `mandrel sync`, and `mandrel update` with no manual re-sync.
  *
  * Flat projection (reverts the #3576 plugin cutover): the plugin command tree
  * (`.claude/plugins/mandrel/`) and the repo-local marketplace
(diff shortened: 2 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4243

**Comparison**: `main` … `story-4243`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4434, opened 2026-07-11
fix: single-story-close: wrong-tree guard intersects stray paths with the Story diff instead of aborting on any main-checkout dirt (#4424)

### LINKED ISSUE #4420: single-story-close: wrong-tree guard false-positives on unrelated concurrent-session dirt in the main checkout
# Summary

The standalone close's wrong-tree guard
(`.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js`,
introduced by #3364) aborts close whenever the **main checkout** has *any*
uncommitted tracked-path change while a per-Story worktree is the active work
tree. It treats every stray tracked file as evidence of the wrong-tree failure
mode (edits intended for the worktree landing in the main checkout, producing
an empty-diff PR).

That inference is too coarse for multi-session operation: uncommitted
tracked-path changes in the main checkout can belong to a **different
concurrent session** and have nothing to do with the Story being closed. When
the Story's worktree already contains a real, committed, non-empty diff, the
failure mode the guard defends against (silent empty-diff PR) is demonstrably
not present — yet the guard still hard-aborts the close.

# Observed false positive (consumer: Beestera/swarm-os, Story #407)

During delivery of Beestera/swarm-os#407 (*Family Portal: Browse Camps page*),
the guard fired and aborted close:

- The 5 stray tracked files it named were **staff-invitation** files
  (`apps/staff/.../invitations*`, `packages/shared/src/staff/invitations*`,
  `apps/staff/src/lib/clerk-invitation-port.ts`) belonging to a concurrent
  session's work that the Story #407 delivery was explicitly instructed not to
  touch.
- All Story #407 edits lived in — and were committed on — the `story-407`
  branch inside `.worktrees/story-407`. The Story diff was real and non-empty;
  the empty-diff failure mode was not present.

Evidence trail:

- Friction comment (the abort): https://github.com/Beestera/swarm-os/issues/407#issuecomment-4936808817
- Progress comment describing the false positive and the workaround: https://github.com/Beestera/swarm-os/issues/407#issuecomment-4936843002
- The delivered PR (real diff, merged): https://github.com/Beestera/swarm-os/pull/408

# Workaround used (and its cost)

Close was re-run with `--cwd` pointed at the Story worktree, putting the runner
in **single-tree mode**, which skips the guard entirely
(`guardApplies` returns `false` when `cwd === worktreePath`). All close phases
completed normally, but this workaround:

1. throws away the guard's protection wholesale rather than refining it, and
2. skipped the close's worktree reap (the flow had to reap manually after the
   merge confirmed).

It also means the documented escape hatch for a false positive is "disable the
guard", which is the wrong ergonomic for a load-bearing check.

# Proposal — distinguish relocated Story work from unrelated dirt

The guard's signal should be *"the Story's work is in the wrong tree"*, not
*"the main checkout is dirty"*. Two complementary refinements (either alone
would have prevented this false positive):

1. **Intersect stray paths with the Story's diff paths.** Compare the main
   checkout's stray tracked paths against the paths changed on the Story
   branch vs the base branch (e.g. `git -C <worktree> diff --name-only
   <base>...HEAD` plus the worktree's uncommitted changes). Abort only when
   the sets overlap — an overlapping path genuinely suggests the same Story
   edit landed in both trees. Disjoint stray paths are another session's
   business: downgrade to a warning (keep the friction comment, don't throw).
2. **Skip the abort when the worktree already has a non-empty committed diff
   vs the base branch.** The failure mode being guarded is a *silent
   empty-diff PR* from an unchanged worktree. If the worktree branch has real
   committed work, that failure mode cannot occur; at most, *some* edits may
   have strayed — which check 1 detects precisely.

Optionally, an operator allowlist (paths/globs known to belong to other
in-flight sessions) could complement the above, but path-intersection alone
seems sufficient and requires no operator input.

The friction-comment posting should stay in all cases — the downgraded
(non-aborting) path is still a useful telemetry signal for concurrent-session
hygiene.

# References

- Guard implementation: `.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js` (#3364)
- Consumer Story where it fired falsely: Beestera/swarm-os#407, PR Beestera/swarm-os#408

### LINKED ISSUE #4424: single-story-close: wrong-tree guard intersects stray paths with the Story diff instead of aborting on any main-checkout dirt
# single-story-close: wrong-tree guard intersects stray paths with the Story diff instead of aborting on any main-checkout dirt

## Context

The standalone close's wrong-tree guard
(`.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js`,
Story #3364) aborts close whenever the **main checkout** has *any*
uncommitted tracked-path change while a per-Story worktree is the active
work tree. That inference is too coarse for multi-session operation:
uncommitted tracked-path changes in the main checkout can belong to a
**different concurrent session** and have nothing to do with the Story
being closed. Framework-gap issue #4420 documents a live false positive
(consumer `Beestera/swarm-os` Story 407): the guard named five stray files
that belonged to another session's work, the Story's own worktree held a
real committed non-empty diff, and the documented workaround (`--cwd` into
the worktree → single-tree mode) disables the guard wholesale and skips
the worktree reap.

The guard's signal should be *"the Story's work is in the wrong tree"*,
not *"the main checkout is dirty"*. Refine `runWrongTreeGuardPhase` so it:

1. **Computes the Story's diff paths** — the union of paths changed on the
   worktree branch vs the base branch
   (`git -C <worktreePath> diff --name-only <baseBranch>...HEAD`) plus the
   worktree's own uncommitted tracked changes
   (`git -C <worktreePath> status --porcelain`, tracked entries only).
2. **Intersects** the main checkout's stray tracked paths with that set.
   - **Overlap non-empty** → genuine wrong-tree signal: post the
     `friction` comment and **throw** (abort close), exactly as today.
   - **Disjoint** → another session's business: post the `friction`
     comment (telemetry for concurrent-session hygiene, with wording that
     says close **proceeded** and names the disjoint stray files) and
     **return without throwing**, logging a `WRONG-TREE` warning progress
     line.
3. **Keeps the existing fail-open probe semantics**: a main-checkout
   status probe failure downgrades to a warning and skips the guard — a
   probe hiccup never blocks an otherwise valid close. A *failed
   Story-diff probe* must not silently convert a would-be abort into a
   pass: if the main checkout has stray tracked paths but the Story-diff
   probe fails, fall back to the current (coarse) abort behavior rather
   than fail-open.

**Empty-diff backstop.** The path-intersection check subsumes issue
#4420's second proposed check ("skip the abort when the worktree already
has a non-empty committed diff") — but an **empty** Story diff-path set
makes every stray path disjoint-by-definition, which would wrongly
downgrade the original #3364 silent empty-diff failure mode. Guard
against that explicitly: when the Story diff-path set is empty (no
committed diff vs base and no uncommitted worktree changes) and the main
checkout has stray tracked paths, keep the abort — that is exactly the
scenario the guard exists for.

The runner call site
(`.agents/scripts/lib/orchestration/single-story-close/runner.js`,
`runPrePushPhases`) already has `baseBranch` and `storyBranch` in scope —
pass them (and whatever else the diff probe needs) into
`runWrongTreeGuardPhase`. Update the existing suite
(`tests/single-story-close-wrong-tree-guard.test.js`) alongside: the
AC-1 abort cases become overlap cases, and new cases cover the disjoint
downgrade, the empty-diff backstop, and the diff-probe-failure fallback.

## Acceptance Criteria

- [ ] `runWrongTreeGuardPhase` aborts (throws + posts an aborted-wording `friction` comment) when at least one main-checkout stray tracked path is also present in the Story's diff-path set (committed diff vs base, or uncommitted tracked change in the worktree).
- [ ] `runWrongTreeGuardPhase` does **not** throw when the Story diff-path set is non-empty and the main checkout's stray tracked paths are fully disjoint from it; it posts a `friction` comment whose wording states close proceeded (not aborted) and names the disjoint stray files, and returns a result that identifies the downgraded outcome (e.g. `{ applied: true, strayFiles: [...], overlap: [] }` or equivalent).
- [ ] Empty-diff backstop: when the Story diff-path set is empty and the main checkout has stray tracked paths, the guard still aborts (the original #3364 silent empty-diff failure mode).
- [ ] The overlap comparison uses repo-relative path equality on both sides (main-checkout porcelain paths vs worktree diff/status paths), so the same file edited in both trees always intersects regardless of which tree the probe ran in.
- [ ] Probe-failure semantics: a main-checkout status probe failure still skips the guard (unchanged); a Story-diff probe failure with stray paths present falls back to the pre-refinement abort behavior instead of silently passing.
- [ ] The runner call site passes the new inputs (base branch, worktree) without changing the behavior of any other close phase, and single-tree mode (`guardApplies` false) remains a no-op.
- [ ] `node --test tests/single-story-close-wrong-tree-guard.test.js` exits 0 with new/updated cases covering: overlap abort, disjoint downgrade (no throw, friction posted, proceeded wording), empty-diff backstop abort, untracked-only main dirt still ignored, and diff-probe-failure fallback abort.
- [ ] `npm run lint` passes and `node .agents/scripts/check-baselines.js` reports no regression on the touched files.

## Out of Scope

- An operator allowlist of paths/globs known to belong to other in-flight
  sessions (issue #4420 names it as optional; path-intersection alone
  resolves the observed false positive).
- Any change to the Epic-path close (`story-close.js`) or to worktree-reap
  behavior — this Story touches only the standalone close's wrong-tree
  guard phase and its runner call site.
- Changing when the guard applies (`guardApplies` single-tree/no-worktree
  semantics stay as-is).

## Notes

- Fixes framework-gap issue #4420 — reference it in commit bodies
  (`refs #4420`) and close it from the PR body (`Closes #4420`).
- Evidence trail for the false positive: Beestera/swarm-os#407
  (friction comment + progress comment) and the merged real-diff PR
  Beestera/swarm-os#408.
- The friction comment stays in **both** outcomes deliberately — the
  downgraded path is telemetry for concurrent-session hygiene, per the
  issue's proposal.

dsj1984:
Closes #4424
Closes #4420

Refines the standalone close's wrong-tree guard (Story #3364) so it intersects the main-checkout stray tracked paths with the Story's own diff-path set (committed diff vs base + uncommitted worktree tracked changes) instead of aborting on any main-checkout dirt. Overlap aborts as before; a non-empty, fully-disjoint Story diff downgrades to a proceed-with-telemetry friction comment (fixes the framework-gap #4420 false positive from concurrent sessions); an empty Story diff keeps the #3364 empty-diff abort backstop; a Story-diff probe failure falls back to the coarse abort. Runner threads baseBranch into the guard.

### DIFF
--- a/.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js
+++ b/.agents/scripts/lib/orchestration/single-story-close/phases/wrong-tree-guard.js
@@ -18,15 +18,33 @@
  * Detection: when a worktree is the active work tree (it exists on disk and is
  * distinct from `cwd`), inspect `git -C <mainCheckout> status --porcelain`. Any
  * **tracked-path** change (modified, staged, deleted, renamed) in the main
- * checkout is the wrong-tree signal. Untracked files (`??`) are ignored — they
- * are scratch artifacts, not relocated Story work, and flagging them would
- * produce false positives on every run.
+ * checkout is the candidate wrong-tree signal. Untracked files (`??`) are
+ * ignored — they are scratch artifacts, not relocated Story work, and flagging
+ * them would produce false positives on every run.
  *
- * The guard runs before the gate chain. On detection it posts a `friction`
- * structured comment naming the stray files and throws, so the operator/agent
- * relocates the edits into the worktree before close proceeds. This is the
- * load-bearing, detection-based fix: it fires regardless of whether the agent
- * ever realized it edited the wrong tree.
+ * Story #4424 — the raw "main checkout is dirty" signal is too coarse for
+ * multi-session operation: uncommitted tracked-path changes in the main
+ * checkout can belong to a **different concurrent session** and have nothing to
+ * do with the Story being closed (framework-gap #4420). The guard therefore
+ * intersects the main-checkout stray tracked paths with the **Story's own
+ * diff-path set** (paths changed on the worktree branch vs the base branch,
+ * plus the worktree's uncommitted tracked changes):
+ *   - **Overlap non-empty** → genuine wrong-tree signal: post a `friction`
+ *     comment and throw (abort close), exactly as before.
+ *   - **Disjoint** (Story diff-path set non-empty, no shared path) → another
+ *     session's business: post a `friction` comment whose wording states close
+ *     PROCEEDED and names the disjoint stray files (telemetry for
+ *     concurrent-session hygiene), then return without throwing.
+ *   - **Empty Story diff-path set** with stray paths → keep the abort (the
+ *     #3364 silent empty-diff backstop; an empty set makes every stray path
+ *     disjoint-by-definition, which must not downgrade the guard).
+ *   - **Story-diff probe failure** with stray paths present → fall back to the
+ *     coarse abort rather than fail-open, so a probe hiccup never converts a
+ *     would-be abort into a silent pass.
+ *
+ * The main-checkout status probe keeps its fail-open semantics: a probe failure
+ * downgrades to a warning and skips the guard — it never blocks an otherwise
+ * valid close.
  */
 
 import path from 'node:path';
@@ -85,6 +103,20 @@ export function collectStrayTrackedPaths(entries) {
 }
 
 /**
+ * Parse `git diff --name-only` output into a list of repo-relative paths.
+ *
+ * @param {string} raw - Raw `git diff --name-only` stdout (may be empty).
+ * @returns {string[]}
+ */
+export function parseDiffNameOnly(raw) {
+  if (typeof raw !== 'string' || raw.trim() === '') return [];
+  return raw
+    .split('\n')
+    .map((line) => line.replace(/\r$/, '').trim())
+    .filter(Boolean);
+}
+
+/**
  * Decide whether the wrong-tree guard applies for this close.
  *
  * The guard only makes sense when a worktree is the active work tree: it must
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4424

**Comparison**: `main` … `story-4424`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4552, opened 2026-07-16
fix(cli): stop sync pruning generated .agents/ files — declare a never-pruned registry so the install ledger survives (#4534)

### LINKED ISSUE #4534: fix(cli): stop sync pruning generated .agents/ files — declare a never-pruned registry so the install ledger survives
## Goal
Stop `mandrel sync` from deleting generated `.agents/` files it did not ship. The prune pass removes every `.agents/` file with no payload counterpart, and `listDestFiles` exempts only `.agents/local/` and `*.local.*` — so `.agents/.install-manifest.json`, the durable install ledger `mandrel uninstall` reads, is destroyed by the first sync after a bootstrap. Because a missing ledger is treated as benign, `mandrel uninstall` then prints "No install ledger found — nothing to uninstall" and exits 0, leaving every bootstrap mutation applied while reporting success. Replace the two hard-coded carve-outs with a declared generated-files registry so this class of bug stops recurring.

## Spec
Filed as: https://github.com/dsj1984/mandrel/issues/4532

### Reproduced, not inferred

Against `main`, with a fixture carrying a bootstrap-written ledger plus two controls:

```
BEFORE .agents/: .install-manifest.json, instructions.local.md, local
runSync result: {"copied":1,"planned":1,"pruned":1,"dryRun":false}
AFTER  .agents/: instructions.local.md, local, scripts

.install-manifest.json : PRUNED  <-- bug
instructions.local.md  : survived (control OK)
local/note.md          : survived (control OK)
```

Both controls surviving isolates the gap to dotfile/generated-file handling rather than the prune pass in general.

### Root cause

`sync.js:262-275` prunes every managed-zone dest file with no payload counterpart. `listDestFiles` (`:144-172`) has exactly two exemptions: the top-level `.agents/local/` zone and `LOCAL_OVERRIDE_RE` (`*.local.*`). The ledger matches neither: it is written by the bootstrap (`install-ledger.js:155`), gitignored by design (`.gitignore:25`; `install-ledger.js:14` calls it "a per-clone install record, not a checked-in source artifact"), and therefore never in `payloadFiles`.

Both modules are individually correct and jointly wrong: the ledger exists to be a generated `.agents/` file that is not payload, and the prune rule is that a non-payload `.agents/` file is stale.

### Impact

`uninstall.js:659` reads the ledger as its single source of truth; `:671-676` treats absence as a **benign no-op and exits 0**. So bootstrap → sync → uninstall (and `sync` runs on every update drift-heal, every `prepare`, and any manual call) silently leaves every reversible mutation applied and every irreversible GitHub-admin mutation unsurfaced, while reporting success. `uninstall.js:778-781` states the contract this violates: "Remove the ledger last so a re-run is a clean no-op and a partial failure mid-reversal still leaves the ledger for a resume." Exactly one thing should delete the ledger — a completed uninstall.

### The fix — a declared registry, not a third special case

Do not just add a third literal to the exemption list. This is the **second** generated-`.agents/`-file class to collide with the prune pass: #4530 specifies `.agents/.mandrel-version` and has to route around this same trap by hand. Introduce a single declared **generated, never-pruned** registry that `listDestFiles` consults, and register both artifacts into it:

- the install ledger, keyed off the exported `LEDGER_RELATIVE_PATH` constant (`install-ledger.js:32`) — never a duplicated string literal;
- the version marker from #4530, if that Story has landed.

Keep the existing `.agents/local/` and `*.local.*` carve-outs as-is — they are a different concept (consumer-authored overrides, not framework-generated artifacts) and conflating them would muddy both.

### Sequencing with #4530

This Story and #4530 both edit the `listDestFiles` exemption set in `lib/cli/sync.js`. #4530 is already `agent::ready` and larger; this Story is marked blocked by it so the registry generalizes an exemption that already exists rather than racing it. **If #4530 has landed when this runs**, fold its marker into the registry and delete the bespoke exemption it added. **If it has not**, build the registry with the ledger alone and leave a clear extension point — do not reach into #4530 scope.

## Changes
- {"path":"lib/cli/sync.js","assumption":"refactors-existing"}
- {"path":"lib/cli/__tests__/sync-prune.test.js","assumption":"refactors-existing"}
- {"path":"tests/e2e/sync-prune.integration.test.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/bootstrap/install-ledger.js","assumption":"exists"}
- {"path":"lib/cli/uninstall.js","assumption":"exists"}

## Acceptance
- [ ] `listDestFiles` in `lib/cli/sync.js` consults a single declared generated-files registry instead of hard-coding individual never-prune paths.
- [ ] The install ledger is registered via the exported `LEDGER_RELATIVE_PATH` constant from `install-ledger.js`, not a duplicated string literal.
- [ ] Running `runSync` twice over a fixture whose `.agents/` carries `.install-manifest.json` leaves the ledger present after both passes (regression test for the reproduced bug).
- [ ] The existing exemptions still hold: `.agents/local/**` and `*.local.*` files survive a sync, and the top-level-only scoping of the `local/` skip is preserved (a nested dir named `local` is still enumerated).
- [ ] A genuinely stale managed-zone file (present in dest, absent from payload, not registered as generated) is STILL pruned — the registry narrows the prune, it does not disable it.
- [ ] The end-to-end chain is proven: after bootstrap → sync, `mandrel uninstall` finds its ledger and reverses the recorded mutations rather than reporting "No install ledger found".
- [ ] If #4530 has landed, its `.agents/.mandrel-version` marker is registered in the same registry and its bespoke exemption is removed; if not, the registry carries the ledger alone with no #4530 scope pulled in.

## Verify
- node --test lib/cli/__tests__/sync-prune.test.js lib/cli/__tests__/sync-local-zone.test.js lib/cli/__tests__/sync.test.js (unit)
- node --test tests/e2e/sync-prune.integration.test.js (e2e)
- node --test tests/cli/uninstall.test.js (unit)
- npm test (unit)
- npm run lint (validate)

## Non-Goals
- Not changing the prune pass itself, the `.agents/local/` zone semantics, the ledger schema, or uninstall behaviour. The question of whether a missing ledger should stay a silent `exit 0` is real but separate — this Story stops the ledger from going missing; it does not re-litigate how uninstall reports an absent one.

dsj1984:
Closes #4534

---

> [!NOTE]
> **Medium Risk**
> Changes sync prune semantics for `.agents/` managed files; incorrect registry entries could leave stale files or delete needed artifacts, but scope is localized to sync and covered by extensive regression tests including real-binary uninstall.
> 
> **Overview**
> **`mandrel sync`** no longer treats bootstrap-generated artifacts as stale during prune. The **`GENERATED_FILES`** registry now exempts **`.agents/.install-manifest.json`** (via **`LEDGER_RELATIVE_PATH`** from `install-ledger.js`) alongside **`.mandrel-version`**, with **`toManagedZoneRelative`** converting generator paths and failing fast if a registry entry is outside `.agents/`.
> 
> **`listDestFiles`** documents and relies on that registry as the single exemption mechanism for framework-generated files (separate from `.agents/local/` and `*.local.*`). Prune behavior otherwise unchanged: unregistered stale files still delete, including lookalike paths and nested `local/` dirs.
> 
> Unit and e2e tests cover ledger survival across syncs/dry-run, registry narrowing (not blanket dotfile skips), and **bootstrap → sync → uninstall** reversing a recorded `.gitignore` merge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07c188de669b1a9103ccd64ada1911f8d80708bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/cli/__tests__/sync-prune.test.js
+++ b/lib/cli/__tests__/sync-prune.test.js
@@ -654,6 +654,54 @@ describe('runSync — the generated-files registry narrows, not disables, prune'
 });
 
 // ---------------------------------------------------------------------------
+// Story #4534 — the local-zone skip stays top-level-only
+// ---------------------------------------------------------------------------
+
+describe('runSync — the local/ prune skip is scoped to the top level', () => {
+  it('still prunes a stale file under a NESTED dir named local/', () => {
+    // The zone exemption is `.agents/local/`, not "any directory named
+    // local". Guards the prune side of that scoping (the copy side is covered
+    // in sync-local-zone.test.js): a nested `local/` is an ordinary
+    // managed-zone directory and its stale files are still enumerated.
+    const nestedLocalFile = path.join(
+      DEST_AGENTS,
+      'rules',
+      'local',
+      'stale-note.md',
+    );
+    const topLevelLocalFile = path.join(DEST_AGENTS, 'local', 'team-notes.md');
+    const fs = makeFs({
+      ...seedPayload(),
+      [nestedLocalFile]: '# nested local — still managed\n',
+      [topLevelLocalFile]: '# consumer-owned — exempt\n',
+    });
+    const cap = makeCapture();
+
+    const result = runSync({
+      argv: [],
+      resolvePackageRoot: resolveToPkg,
+      fs,
+      cwd: () => PROJECT,
+      write: cap.write,
+      writeErr: cap.writeErr,
+      exit: cap.exit,
+    });
+
+    assert.equal(
+      fs.files.has(nestedLocalFile),
+      false,
+      'a stale file under a nested local/ dir must still be pruned',
+    );
+    assert.equal(
+      fs.files.has(topLevelLocalFile),
+      true,
+      'the top-level .agents/local/ zone stays exempt',
+    );
+    assert.equal(result.pruned, 1);
+  });
+});
+
+// ---------------------------------------------------------------------------
 // Story #4534 — the registry is the SSOT the generators key off
 // ---------------------------------------------------------------------------
 
--- a/lib/cli/__tests__/sync-prune.test.js
+++ b/lib/cli/__tests__/sync-prune.test.js
@@ -18,6 +18,16 @@
  *      the tree byte-identical and prunes nothing.
  *   6. `result.pruned` reports the number of deleted stale files.
  *
+ * Extended by Story #4534 (the generated, never-pruned registry):
+ *   7. A bootstrap-written install ledger survives two consecutive syncs —
+ *      the reproduced bug (#4532) that made `mandrel uninstall` silently
+ *      reverse nothing.
+ *   8. The registry NARROWS the prune rather than disabling it: an
+ *      unregistered stale dotfile, and a lookalike basename at a
+ *      non-registered path, are both still pruned.
+ *   9. The registry is keyed off each generator's own exported constant
+ *      (`LEDGER_RELATIVE_PATH`), never a duplicated literal.
+ *
  * Tier: unit (testing-standards § Unit). All filesystem I/O is mocked via
  * the in-memory fs fake used by sync.test.js.
  *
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4534

**Comparison**: `main` … `story-4534`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4567, opened 2026-07-16
fix(plan): make the fan-out deletion gate count importers, not basename word matches (#4547)

### LINKED ISSUE #4547: fix(plan): make the fan-out deletion gate count importers, not basename word matches
## Goal
Make the plan-persist fan-out gate measure coupling instead of vocabulary. The gate exists to stop a Story from casually deleting a widely-imported module, but its counter greps the deleted file basename as a bare word across the whole tree — so a module named notification or options trips it regardless of who imports it, while the number it reports bears no relation to the risk it is protecting against.

## Spec
Observed while planning a residue sweep on 2026-07-16. The counter takes the deleted path basename, strips the extension, and runs a repo-wide grep for that stem, counting distinct matching files. For the post-merge notification phase it reported 59 call sites; only 3 files in the tree reference that module at all, and the other matches were any file containing the word notification — the SDLC narrative, unrelated schemas, the agentrc reference. The worktree-reap phase reported 18 for a similar reason: a live close-path phase shares its basename. In both cases the deleted modules had zero production importers, verified by path-qualified grep.

The consequence is worse than a wrong number. The gate is fail-closed and its message tells the operator to either split the deletion across Stories or re-run with an override. When the count is noise, splitting is not a real option — there is nothing to migrate — so the only path forward is the override. An operator who learns that the override is the normal response to this gate will pass it on the day the count is real. A gate that cries wolf is worse than no gate, because it manufactures the habit that defeats it.

The fix is to probe references, not vocabulary. Resolve the deleted path against import and require statements — a path-qualified suffix match is sufficient and cheap, and the repo already greps at a base ref for the freshness gate, so the probe shape exists. Two secondary defects belong in the same pass: the counter returns zero for any basename stem shorter than three characters without probing at all, which silently under-reports rather than over-reports; and the reported figure is currently unauditable, so an operator cannot tell a real 30 from a noisy 59 without reproducing the grep by hand. Name the referencing files, or name the probe.

Keep the gate. Its intent is sound and this Story does not weaken it: a genuinely wide-coupling deletion must still fail closed at the threshold. The change is that the number it fires on should mean what the message says it means.

## Changes
- {"path":".agents/scripts/lib/orchestration/plan-persist/persist-helpers.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/orchestration/plan-persist/fan-out-gate.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/orchestration/ticket-validator-conflicts.js","assumption":"refactors-existing"}
- {"path":"tests/scripts/plan-persist.flat-stories.test.js","assumption":"refactors-existing"}
- {"path":"tests/ticket-validation.test.js","assumption":"refactors-existing"}

## Acceptance
- [ ] The fan-out gate counts files that actually reference the deleted module — by import or require of its path — rather than files containing its basename as a bare word.
- [ ] A deletion of a module with a generic basename that no code imports reports zero call sites, where the current counter reports dozens of unrelated matches.
- [ ] A deletion of a module that many files genuinely import still trips the gate at the configured threshold.
- [ ] The reported count is reproducible by an operator: the gate names the referencing files, or the exact probe that produced the number, so the figure can be checked rather than trusted.
- [ ] A module whose basename is shorter than the current minimum stem length is probed correctly rather than silently reported as zero call sites.
- [ ] The gate message distinguishes a genuine wide-coupling deletion from a rename-shaped one, so the suggested remedy matches the situation.

## Verify
- node --test tests/scripts/plan-persist.flat-stories.test.js (unit)
- node --test tests/ticket-validation.test.js (unit)
- npm run lint (validate)
- npm test (unit)

## Non-Goals
- The threshold value itself, and whether it should be operator-tunable.
- The other persist gate defects — the base-branch config path, the subject-prefix blindness to string bodies, and the parse-failure that degrades into a misleading freshness miss. Those are covered by the persist-integrity Story, which this one follows because both edit the same helper module.
- The dead-exports ratchet and its test-only-importer blindness, which is a separate signal with its own Story.

dsj1984:
Closes #4547

---

> [!NOTE]
> **Medium Risk**
> Changes plan-persist validation that can block or warn on deletions; behavior is more accurate but relative-import-only probing may under-count alias/bare imports (documented).
> 
> **Overview**
> **Replaces the large-fan-out deletion probe** so it measures actual import/require coupling instead of grepping the deleted file’s basename stem as a bare word (which inflated counts from docs/schemas and skipped short stems without probing).
> 
> `makeDefaultFanOutCounter` now runs a two-stage check: `git grep -n -E` for candidate import lines, then resolves **relative** specifiers per importer file. It returns `{ count, files, probe }` with a **shell-quotable** `git grep` command operators can paste to audit the number.
> 
> **Fan-out findings and persist errors** carry full importer lists (not truncated), optional probe text, and **context-specific remedies**: rename-shaped deletes (same basename `creates` elsewhere) advise repointing importers in one Story instead of always pushing a multi-Story split. Legacy counters that return a plain number still work but omit evidence.
> 
> Extensive unit/integration tests cover false-positive basenames, short stems, index modules, self-references, and rename detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13f53e0e5d218e540c8d189640f1e6ab010fadf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4547

**Comparison**: `main` … `story-4547`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4570, opened 2026-07-16
feat(plan): mirror sibling depends_on into native GitHub blocked_by edges at persist (#4544)

### LINKED ISSUE #4544: feat(plan): mirror sibling depends_on into native GitHub blocked_by edges at persist
## Goal
Make plan-authored Story ordering durable as data rather than prose. The helper that translates depends_on edges into native GitHub blocked_by dependencies already exists, is tested, is idempotent, and is non-fatal by contract — and has had zero production callers since the v2 cutover. Wire it into story creation so sibling order is visible in the GitHub UI and readable by the delivery resolver without parsing body text.

## Spec
The writer is already written. applyBlockedByDependencies translates authored depends_on slugs into native blocked_by edges with bounded concurrency, idempotent re-application, and a deliberately non-fatal failure mode. Its only importer is its own test file — the v1 decomposition seam that called it was deleted in the cutover, and v2 story creation never picked it up.

Two obstacles the wiring must solve rather than assume. First, a signature gap: story creation receives only the provider, while the writer needs the API handle plus owner and repo — reaching those means touching a private-by-convention provider field from an orchestration module, which is a layering decision, not a wire-up. Pass what it needs through an explicit interface. Second, and more dangerous, a silent shape mismatch: story creation builds its slug-to-id lookup as a Map, while the writer indexes it as a plain object. Property access on a Map yields undefined, so the writer would warn once per story, skip every edge, and return zero edges added — and because the contract is non-fatal, persist would report success while writing nothing. That is why the acceptance asserts an exact edge count rather than mere completion.

Non-fatal is right here, in contrast to the read path. A missing native edge is cosmetic: ordering still survives in the blocked by footer that story creation already serializes and the dependency parser already reads. A dropped read-side edge, by contrast, removes a real dispatch gate — which is why the resolver work treats a failed read as a hard error. Keep the asymmetry deliberate and documented.

This is carved out of the deliver resolution work on a real seam: the resolver needs only to READ edges (operator-set ones included, which is how cross-run order is expressed today), and it works without this Story. This adds the write side so plan-created order stops depending on prose.

## Changes
- {"path":".agents/scripts/providers/github/blocked-by-add.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/orchestration/plan-persist/story-ops.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/providers/github.js","assumption":"refactors-existing"}
- {"path":"tests/providers/github/blocked-by-add.test.js","assumption":"refactors-existing"}
- {"path":"tests/lib/orchestration/plan-persist-story-ops.test.js","assumption":"refactors-existing"}
- {"path":"baselines/dead-exports.json","assumption":"refactors-existing"}

## Acceptance
- [ ] When a plan creates sibling Stories with authored depends_on edges, each edge is mirrored into a native GitHub blocked_by dependency on the created issues, so the ordering is visible in the GitHub UI and readable by the resolver without parsing body prose.
- [ ] A test asserts the exact number of edges added for a multi-Story plan carrying one dependency, so a wiring mistake that adds zero edges fails rather than reporting success.
- [ ] Applying the edges is idempotent: persisting a plan whose native edges already exist adds no duplicates and reports zero new edges.
- [ ] A dependencies-API rejection is non-fatal: the persist completes, the ordering survives in the body footer, and the failure is reported rather than swallowed silently.
- [ ] The edge writer receives the identifiers it needs through an explicit interface rather than reaching into provider internals from the orchestration layer.

## Verify
- node --test tests/providers/github/blocked-by-add.test.js (unit)
- node --test tests/lib/orchestration/plan-persist-story-ops.test.js (unit)
- node --test tests/scripts/plan-persist.flat-stories.test.js (unit)
- node .agents/scripts/check-dead-exports.js (validate)
- npm run lint (validate)
- npm test (unit)

## Non-Goals
- Reading native edges during delivery resolution — that is the /deliver ids-only Story, which must land first so the read semantics (issue numbers, fail-loud, cross-repo rejection) are already settled.
- Cross-repository dependency edges.

dsj1984:
Closes #4544

---

> [!NOTE]
> **Medium Risk**
> Changes plan-persist behavior and GitHub dependency writes, but failures are non-fatal and ordering remains in issue bodies; mis-wiring could silently skip native edges until caught by tests.
> 
> **Overview**
> After all Story issues exist (including on a resumed persist), **plan persist** now mirrors authored sibling `depends_on` into **native GitHub `blocked_by` dependency edges** via `applyBlockedByDependencies`, while keeping the existing `blocked by #N` body footers as the fallback ordering source.
> 
> **`GitHubProvider`** gains **`getDependencyWriteContext()`** (`gh`, `owner`, `repo`) so orchestration can call the dependencies REST API without reaching into private `_gh`. Mirroring is **non-fatal**: API or wiring failures log and return counts; persist still completes. **`createStoryIssues`** now also returns a **`dependencyEdges`** summary (or `null` when skipped).
> 
> Tests cover POST shape (dependent issue, blocker database id), idempotency, API failure, missing interface, no-edge plans, and resumed/adopted runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4dc05891034c271ca79385ccdac279b127b50c75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/orchestration/plan-persist/story-ops.js
+++ b/.agents/scripts/lib/orchestration/plan-persist/story-ops.js
@@ -16,6 +16,7 @@
  */
 
 import { createHash } from 'node:crypto';
+import { applyBlockedByDependencies } from '../../../providers/github/blocked-by-add.js';
 import { Logger } from '../../Logger.js';
 import { AGENT_LABELS, TYPE_LABELS } from '../../label-constants.js';
 import {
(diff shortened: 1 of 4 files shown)


dsj1984:
## 🔬 Code Review — Story #4544

**Comparison**: `main` … `story-4544`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4610, opened 2026-07-17
fix(story-close): close #4593's single-enumeration leak and root-cause the un-fired MI gate

### LINKED ISSUE #4593: Compute the Story change-set once and thread it through delivery
## Goal
One Story delivery enumerates `git diff --name-only <base>...<head>` three to four times; compute it once so ceremony routing, review depth, and the local-lens pass can never disagree about what changed.

## Spec
Current enumeration sites (verified 2026-07-17): the ceremony derivation step in `helpers/deliver-story.md`; `review-depth.js`'s `listChangedFiles` invoked from `runCodeReview` in `.agents/scripts/lib/orchestration/code-review.js`; the local-lens `enumerateChangedFiles` in `.agents/scripts/lib/orchestration/single-story-close/phases/code-review.js`; and each fresh acceptance-critic re-derives the diff per `.agents/agents/acceptance-critic.md`. `.agents/scripts/lib/audit-suite/selector.js#selectAudits` already accepts an injected `changedFiles` list.

Approach:

1. **One enumerator.** Extract a single `computeChangeSet(baseRef, headRef)` into a small shared module (e.g. `.agents/scripts/lib/orchestration/change-set.js`): argv-based git call, returns a sorted, de-duplicated file list plus the refs it was computed against.
2. **Injection over recomputation.** `runStoryReviewCore` computes the change-set once at entry and passes it to both `runLocalLensReview` and `runCodeReview`; `review-depth.js` and the local-lens phase accept the injected list and only fall back to self-enumeration when none is supplied (preserves standalone CLI use).
3. **Workflow threading.** `helpers/deliver-story.md` Step 1a computes the change-set once (same module via a tiny CLI entry or inline `node -e` is acceptable — pick the cleanest seam) and reuses it for ceremony derivation; `.agents/agents/acceptance-critic.md` receives the file list in its spawn context instead of instructing a fresh `git diff`, keeping the critic's evidence anchored to the same set the router saw.
4. **Consistency invariant.** Because ceremony level and review depth both flow from `deriveChangeLevel`, feeding them one list closes the window where commits landing between steps let the two derive from different diffs.

## Changes
- {"path":".agents/scripts/lib/orchestration/change-set.js","assumption":"creates"}
- {"path":".agents/scripts/lib/orchestration/code-review.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/orchestration/review-depth.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/lib/orchestration/single-story-close/phases/code-review.js","assumption":"refactors-existing"}
- {"path":".agents/workflows/helpers/deliver-story.md","assumption":"refactors-existing"}
- {"path":".agents/agents/acceptance-critic.md","assumption":"refactors-existing"}
- {"path":"tests/change-set.test.js","assumption":"creates"}
- {"path":"tests/review-depth.test.js","assumption":"refactors-existing"}

## Acceptance
- [ ] A shared change-set module exists and `node --test tests/change-set.test.js` passes, covering sorted/deduplicated output and ref reporting (unit test)
- [ ] `runStoryReviewCore` passes one injected change-set to both the local-lens phase and `runCodeReview`; a test asserts the git enumerator is invoked exactly once per close run (unit test with seam/spy)
- [ ] `review-depth.js` and the local-lens phase accept an injected `changedFiles` list and fall back to self-enumeration only when it is absent (asserted by unit test)
- [ ] `grep -n "git diff" .agents/agents/acceptance-critic.md` exits 1 — the critic boot context receives the file list from its caller instead of instructing a fresh diff
- [ ] Existing review-path suites still pass unmodified in behavior: `node --test tests/story-close-review.test.js tests/review-depth.test.js tests/code-review.test.js`

## Verify
- node --test tests/change-set.test.js (unit)
- node --test tests/story-close-review.test.js tests/review-depth.test.js tests/code-review.test.js (unit)
- npm run lint (validate)

## References
- {"path":".agents/scripts/lib/audit-suite/selector.js","assumption":"exists"}
- {"path":".agents/scripts/lib/orchestration/ceremony-routing.js","assumption":"exists"}

## Non-Goals
- No change to lens selection, review-depth thresholds, or ceremony-routing policy — this is plumbing, not policy.
- No caching across close invocations; the change-set is computed per run, once.
- `selectAudits`'s existing injection contract is consumed, not modified.

### LINKED ISSUE #4603: Close #4593's single-enumeration leak on the failure path and root-cause the MI gate that let its ratchet violation land
## Goal
#4593's "enumerated exactly once" invariant is false on the unenumerable path, the test that should catch it is neutralized by its own stub, and the same Story landed a -5.86 maintainability ratchet violation that a must-fix gate failed to fire on.

## Spec
Current state (verified 2026-07-17 by the post-delivery audit-lens walk over `cf90bea7..origin/main`, re-verified by hand).

**Defect A — the null/undefined conflation.** `runStoryReviewCore` in `.agents/scripts/lib/orchestration/story-close/phases/code-review.js` asserts it is "the one enumeration per close run … none of them re-derives the diff". When `computeChangeSet` fails, `changeSet.files` is `null`; `runLocalLensReview` in the same file discriminates with `Array.isArray(injectedChangedFiles)`, collapsing the load-bearing `null` ("already tried, unenumerable") and `undefined` ("nobody enumerated") states, and re-spawns `git diff`. The sibling `resolveInjectedChangedFiles` in `.agents/scripts/lib/orchestration/code-review.js` gets this right by checking `=== undefined` first. Outcome is benign today (`[]` either way, the retry fails identically) — the defect is that the docstring asserts an invariant the code does not hold, and that is the kind of comment the next reader trusts.

**Defect B — the test is neutralized by its stub.** `tests/audit-suite/story-scope-lens-review.test.js` has a case named for injecting an explicit null when the diff is unenumerable, which asserts the git enumerator ran once while passing a stubbed `runLocalLensReviewFn` — so the real collaborator's null branch never executes. A second case passes no `changedFiles` key at all, testing `undefined`, never `null`. The suite reports the invariant as held when production does not hold it.

**Defect C — an un-gated ratchet violation, and the blind spot behind it.** `node .agents/scripts/quality-preview.js --changed-since cf90bea7` reports `MI regressions=1 · gate exits: mi=1`, a **-5.86** delta on `.agents/scripts/lib/orchestration/story-close/phases/code-review.js` (current ~90.61 vs baseline 96.469) against a configured absolute tolerance of 0.5 (`.agentrc.json` → `delivery.quality.gates.maintainability`). #4593 owns it alone and bumped no baseline. Absolute MI is far above the floor of 70, so this is a ratchet violation rather than unmaintainable code — but a must-fix gate did not fire on merge and the erosion is now the de-facto baseline.

**Root-cause C before fixing the number.** Two candidate blind spots, both observed: (1) the `quality:preview` npm script hardcodes `quality-preview.js --changed-since HEAD`, so every `npm run quality:preview -- --changed-since <base>` appends a second flag and silently compares against HEAD instead — a reviewer checking a branch gets a false green. (2) `baselines/maintainability.json` carries two rows whose paths collide by suffix: `.agents/scripts/lib/orchestration/single-story-close/phases/code-review.js` (mi 90.602) and `.agents/scripts/lib/orchestration/story-close/phases/code-review.js` (mi 96.469) — the former *ends with* the latter, so any suffix or substring baseline lookup can read the wrong row, and 90.602 is suspiciously close to the post-change 90.61. Verify whether the resolver is exact-match before concluding; if it is, find the real reason the gate did not fire at commit time. Fix the blind spot, then bring the file back within tolerance by extraction (never by bumping the baseline).

## Changes
- {"path":".agents/scripts/lib/orchestration/story-close/phases/code-review.js","assumption":"refactors-existing"}
- {"path":"tests/audit-suite/story-scope-lens-review.test.js","assumption":"refactors-existing"}
- {"path":".agents/scripts/quality-preview.js","assumption":"refactors-existing"}
- {"path":"package.json","assumption":"refactors-existing"}

## Acceptance
- [ ] `runLocalLensReview` distinguishes injected `null` (degrade, do not re-enumerate) from `undefined` (self-enumerate), matching `resolveInjectedChangedFiles`'s contract (asserted by unit test)
- [ ] The unenumerable-diff test exercises the REAL `runLocalLensReview` collaborator (not a stub) and fails if the null branch re-spawns git — verified by reverting the fix and observing the test go red
- [ ] Running the maintainability gate against a branch base no longer silently compares against HEAD: an invocation passing an explicit base reports that base's delta (asserted by test or by a documented, checked invocation)
- [ ] The root cause of the un-fired gate is named in the PR body with evidence, and the suffix-collision hypothesis for `baselines/maintainability.json` is either confirmed with a fix or ruled out with evidence
- [ ] `node .agents/scripts/quality-preview.js --changed-since <this Story's base>` reports `MI regressions=0` with no baseline bump to `baselines/maintainability.json` (`git diff --stat <base> -- baselines/maintainability.json` is empty)

## Verify
- node --test tests/audit-suite/story-scope-lens-review.test.js (unit)
- node --test tests/change-set.test.js tests/code-review.test.js (unit)
- node .agents/scripts/quality-preview.js --changed-since main (validate)
- npm run lint (validate)

## References
- {"path":".agents/scripts/lib/orchestration/code-review.js","assumption":"exists"}
- {"path":".agents/scripts/lib/orchestration/change-set.js","assumption":"exists"}
- {"path":"baselines/maintainability.json","assumption":"exists"}
- {"path":".agentrc.json","assumption":"exists"}

## Non-Goals
- Do not bump `baselines/maintainability.json` to absorb the -5.86; the ratchet is the point. Extract to bring the file back within tolerance.
- No change to lens selection, review-depth thresholds, or ceremony-routing policy.
- `.agents/scripts/lib/orchestration/change-set.js`'s public contract (`files: null` vs `[]`) is correct and stays as-is; the consumers are what need fixing.

dsj1984:
Closes #4603

#4593 shipped three coupled problems: a documented invariant its code did not hold, a test that proved the invariant *because its own stub neutralized it*, and a -5.86 maintainability ratchet violation that a must-fix gate never fired on. This closes all three and root-causes the gate.

## Defect A — the null/undefined conflation

`runStoryReviewCore` asserts *"the one enumeration per close run … none of them re-derives the diff"*. `runLocalLensReview` discriminated the injected change set with `Array.isArray()`, collapsing two load-bearing states:

- `null` — the spine already tried; the diff is **unenumerable**.
- `undefined` — **nobody** enumerated (standalone callers).

Both fell to the same `else`, so the failure path re-spawned `git`. The sibling `resolveInjectedChangedFiles` in `orchestration/code-review.js` already gets this right by checking `=== undefined` first.

Fixed by `resolveLensChangeSet`, mirroring that contract: `undefined` → self-enumerate; `null` → degrade to the fail-safe empty roster **without retrying**; array → verbatim.

The outcome was benign today (`[]` either way, the retry fails identically). The defect is that the docstring asserted an invariant the code did not hold — the kind of comment the next reader trusts.

## Defect B — the test was neutralized by its own stub

The case named for the null branch passed a **stubbed** `runLocalLensReviewFn`, so the real collaborator's null branch never executed; a second case passed no `changedFiles` key at all (testing `undefined`, never `null`). The suite reported the invariant as held while production broke it.

The spy now **delegates to the real `runLocalLensReview`**. Proved red before the fix:

```
✖ runStoryReviewCore injects an explicit null when the diff is unenumerable
  AssertionError: an unenumerable diff must be enumerated ONCE — the real lens
  pass must not retry git on the null branch, got 2 enumeration(s)
✖ runLocalLensReview degrades on an injected null without re-enumerating
  AssertionError: an injected null must NOT re-spawn git
ℹ pass 16 · fail 3
```

Restored → 19/19 green. The delegation bites because `review-core.js` threads the test's `gitSpawnFn` into the lens args, so a retry is actually counted.

## Defect C — root cause of the un-fired MI gate (with evidence)

The gate did not fire for a **layered** reason. All three layers are reproduced below.

### 1. The only gate that scores source against the MI baseline is a git hook — and it was silently dead

`check-baselines.js` (the close/CI required check) compares **the committed baseline file at head vs at base** — it never re-scores the source tree. An un-refreshed `baselines/maintainability.json` is therefore green *by construction*; it can only catch you if you regenerate the baseline. The **pre-commit hook** (`quality-preview.js --staged`) is the only thing that scores real source against the baseline.

That hook does not run. `.husky/_` is absent and `core.hooksPath` is unset in both the main checkout and the worktrees, so husky v9 was never initialized and **no** git hook fires — not pre-commit (MI/CRAP), not commit-msg (commitlint), not pre-push. Proven by staging a change the gate reports as `mi=1` and committing it successfully:

```
$ node .agents/scripts/quality-preview.js --staged
| .../story-close/phases/code-review.js | -3.69 | ... |
Totals: MI regressions=1   [quality:preview] gate exits: mi=1
$ git commit -m "probe"      # ← succeeded anyway
```

This is machine state (`npm install` re-runs `prepare` → `husky`), not repo code, so it is **not** fixed here — but it is the proximate cause and it means the repo's sole source-scoring MI gate is one un-run `prepare` away from silence on any clone. Follow-up filed for durable CI enforcement.

### 2. The manual escape hatch printed a false green

`npm run quality:preview -- --changed-since <base>` silently compared against `HEAD`. The cause is sharper than a parsing bug: **npm appends `--` passthrough args to the LAST command of a compound `A && B` script only.** Proven directly:

```
script: node echo.js FIRST && node echo.js SECOND
$ npm run probe -- --changed-since MYBASE
FIRST  received: []
SECOND received: ["--changed-since","MYBASE"]
```

`quality:preview` ran `quality-preview.js` **first**, so an operator's base landed on `check-dead-exports.js` (which ignores unknown flags) and the MI gate **never saw it** — it scored against its own hardcoded `--changed-since HEAD` and reported a green for a branch it had never compared. That is how the -5.86 cleared a hand-check.

Fixed by putting the flag-consuming gate last, pinned by a test that fails if another command is ever appended after it. `parseChangedSinceArg` is also hardened to last-occurrence-wins (the conventional CLI semantic, and the shape the canonical single-command `QUALITY_NPM_SCRIPTS` alias relies on).

Verified:

```
$ npm run quality:preview -- --changed-since 0875b6af
scope=diff ref=0875b6af      # was: ref=HEAD
```

### 3. The suffix-collision hypothesis — RULED OUT, with evidence

`baselines/maintainability.json` carries two rows that collide on the shared tail `story-close/phases/code-review.js` (because `single-story-close` itself ends with `story-close`):

| row | path | mi |
| --- | --- | --- |
| A | `…/orchestration/single-story-close/phases/code-review.js` | 90.602 |
| B | `…/orchestration/story-close/phases/code-review.js` | 96.469 |

Both satisfy `endsWith(tail)`, so a suffix-keyed lookup would be ambiguous and a `find()`-style resolver would return **A (90.602)** — which is ~B's post-#4593 score (90.61), presenting a real -5.86 regression as a **~0.008 delta** that sails through the 0.5 tolerance. That is exactly the masking the hypothesis proposed, and it is why the numbers looked suspicious.

**It does not happen.** Resolution is a plain object-key lookup — `projectMaintainabilityEnvelopeToFlat` builds `flat[row.path]`, and `preview-gates.js#compareScores` reads `baseline[file]`. Exact match; no `endsWith`, no substring, no basename fallback anywhere under `lib/baselines/`. Demonstrated:

```
exact lookup: …/single-story-close/phases/code-review.js -> 90.602
exact lookup: …/story-close/phases/code-review.js        -> 96.469
```

One correction to the hypothesis as filed: "the former **ends with** the latter" is **literally false** for the full paths (`A.endsWith(B) === false` — B carries the shared `.agents/scripts/lib/orchestration/` prefix). The collision is **tail-only**, which is the sole shape a fuzzy resolver could confuse. `tests/baselines/maintainability-baseline-exact-match.test.js` records this ruling-out as executable evidence and pins that premise explicitly, so the fixture cannot drift into vacuity and a future move to fuzzy path resolution fails loudly instead of re-blinding the ratchet.

## The ratchet, repaid by extraction (no baseline bump)

Per the Non-Goals, the -5.86 is repaid by **extraction**, never by absorbing it into the baseline:

- `review-core.js` — the shared spine both close paths call `runCodeReview` through. It belongs to neither path, so it no longer lives inside one path's phase file.
- `local-lens-review.js` — the shift-left local-lens pass (selection + materialization).
- `code-review.js` — now just the Epic-attached phase entry point, with `runStoryCodeReview` decomposed into named helpers (`invokeReviewCore`, `emitCriticalBlock`, `resolveSeverity`, …).

`.../story-close/phases/code-review.js`: **90.61 → 102.70** against a 96.469 baseline (tolerance 0.5).

Worth recording, because it is counter-intuitive and cost real time: escomplex's module maintainability tracks the **method average**, not the module total — so *removing* a large, simple function made the file score **worse** (90.61 → 92.37 → 89.42 → 85.89 as extraction proceeded), and what actually fixed it was decomposing the remaining 55-line function into small named helpers. Extraction alone is not a reliable MI lever here; extraction **plus** decomposition is.

Also worth noting: the 96.469 baseline was already stale before #4593 — the pre-#4593 file scores 92.78 (-3.69). #4593's own contribution was ~-2.17. The file now clears the original baseline on its own merits.

## Verification

- `git diff --stat 0875b6af -- baselines/maintainability.json` → **empty** (no baseline bump)
- `node .agents/scripts/quality-preview.js --changed-since 0875b6af` → `MI regressions=0 · CRAP violations=0`
- `node --test tests/audit-suite/story-scope-lens-review.test.js` → 19/19
- `npm run lint` → 0 · `node .agents/scripts/check-baselines.js` → 0
- Full suite: **7772 pass, 0 fail**

Non-Goals respected: no change to lens selection, review-depth thresholds, or ceremony-routing policy; `change-set.js`'s `files: null` vs `[]` public contract is untouched — the consumers were what needed fixing.

---

> [!NOTE]
> **Medium Risk**
> Touches story-close review orchestration and change-set injection on failure paths; quality-preview ordering affects how maintainability gates run locally, though behavior is test-pinned.
> 
> **Overview**
> **Story-close review** is split so `review-core.js` owns the shared `runStoryReviewCore` spine (one `computeChangeSet` per close) and `local-lens-review.js` owns lens selection/materialization; epic-attached `code-review.js` is reduced to phase wiring plus small helpers. The standalone close path imports the spine from `review-core.js`.
> 
> **Enumeration bug (#4593):** `runLocalLensReview` now uses `resolveLensChangeSet` so injected `null` means “diff already failed—degrade to `[]` without re-running git,” `undefined` triggers self-enumeration, and arrays pass through. Tests delegate to the real lens pass on the unenumerable path instead of stubbing it away.
> 
> **MI gate escape hatch:** `quality:preview` runs `check-dead-exports` first and `quality-preview.js` last so `npm run … -- --changed-since <base>` reaches the MI gate; `parseChangedSinceArg` uses **last** `--changed-since` when the flag repeats. New tests lock script order and baseline path **exact-match** (ruling out suffix collision between the two `code-review.js` baseline rows).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd90393aa6547c3c1b64c9744b1292e4ddbbb5be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4603

**Comparison**: `main` … `story-4603`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

#### 🟢 Lint check passed with 2 warning(s) _[lint]_

Scoped lint reported 2 warning(s) on the changed surface. Treat as suggestions.

PR #4661, opened 2026-07-21
fix(feedback-loop): repair the graduator idempotency probe — an HTML-comment-wrapped search query never matches, so every rollup re-files (#4657)

### LINKED ISSUE #4657: fix(feedback-loop): repair the graduator idempotency probe — an HTML-comment-wrapped search query never matches, so every rollup re-files
## Goal
Make the graduator idempotency probe actually find already-filed follow-ups, so a recurring finding files one issue instead of one per rollup invocation.

## Slicing
1. Normalize the `gh search issues` query in `probeMarkerExists` and correct the marker-indexability doc claims.
2. Add the strongly-consistent label-scoped confirmation on the would-file path, ahead of `createFollowUpIssue`.
3. Thread a per-invocation filed-marker memo through the graduate walk and across the retro graduator's two source buckets.
4. Regression tests: wrapped-marker query never reaches `gh`, a same-window duplicate is caught by the strong read, and both graduators keep their existing already-filed and legacy-marker behaviour.

## Spec
Current state (verified 2026-07-20): `probeMarkerExists` passes the whole HTML-comment marker as the `gh search issues` query — `gh search issues "<!-- retro-proposal-followup: epic-N-<fp> -->" --repo …`. That query form never matches, so the probe has effectively always returned `false` and `resolveAlreadyFiled` has always answered "not filed".

Measured against this repo on 2026-07-20, all four forms querying an issue whose body provably contains the marker text:

| Query | Result |
| --- | --- |
| `682ca6649894afd4` (bare fingerprint) | matches |
| `epic-648-682ca6649894afd4` | matches |
| `retro-proposal-followup: epic-648-682ca6649894afd4` | matches |
| `<!-- retro-proposal-followup: epic-648-682ca6649894afd4 -->` | **no match** |

A dedicated probe issue (dsj1984/mandrel#4656, since closed) carried the marker **only** inside an HTML comment. Its bare token matched within 20s and kept matching; the wrapped form returned empty at 20/40/60/80/100/120s. Two conclusions the reported diagnosis did not have: GitHub **does** index HTML-comment text, and the search index caught up in under 20 seconds. Index latency is therefore a narrow residual, not the cause — the `<` / `>` delimiters are.

A second correction: the reported "cross-scope marker collision" is not a collision. `executeFollowUpRollup` passes `epicId: primaryId` (the run's first Story id) and `captureStoryFollowUps` passes the same Story id, so the two scopes mint an identical marker by construction. That is the intended identity — the dedup was supposed to catch it and could not.

**Approach — fix the query, then close the residual window.**

1. **Normalize the search query in `probeMarkerExists`.** Strip the `<!--` / `-->` delimiters and trim before spawning `gh search issues`; leave the caller-facing marker argument unchanged so both graduators and both marker kinds (content-hash and legacy ordinal) are fixed in one place. Callers still pass the full marker — the normalization is the probe's own concern. Repair the three doc comments in `graduator-core.js` and `audit-results-graduator.js` that assert the wrapped marker "stays indexable via `gh search`".

2. **Confirm strongly before filing.** The measured sub-20s index window is exactly the 16-second gap between the two byte-identical duplicates. So when the corrected search probe reports "not found", confirm against a strongly-consistent read before creating: `gh issue list --repo <routed> --state all --json number,body` narrowed by the follow-up's own labels, matching the marker as a substring of the returned bodies. Label-scoping is sound because the same `spec.buildFollowUp` call that writes the marker also supplies the labels, so the two agree by construction; resolve the labels before the dedup decision rather than after. The extra list call happens only on the would-file path, never on the already-filed path.

3. **Memoize within the walk.** `graduateRetroProposals` calls `graduate()` once per source bucket, so a per-invocation set of markers filed so far short-circuits a repeat without any spawn. Cheap, and it keeps the two buckets from racing each other.

Failure posture is unchanged and deliberate: a probe that cannot decide still degrades toward filing rather than swallowing the finding. What changes is that "cannot decide" stops being the only outcome.

## Changes
- `.agents/scripts/lib/feedback-loop/graduator-core.js` — refactors-existing
- `.agents/scripts/lib/feedback-loop/retro-proposals-graduator.js` — refactors-existing
- `.agents/scripts/lib/feedback-loop/audit-results-graduator.js` — refactors-existing
- `tests/lib/feedback-loop/graduator-core.test.js` — refactors-existing
- `tests/lib/retro-proposals-graduator.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: The probe never sends comment delimiters to the search index: `probeMarkerExists` called with `<!-- retro-proposal-followup: epic-1-abc -->` spawns `gh` with an argument list containing no `<!--` and no `-->` substring.
- [ ] AC-2: The probe still identifies the marker it was given: a stubbed `gh search issues` that matches only on the undelimited marker text returns `alreadyFiled: true` through `resolveAlreadyFiled`.
- [ ] AC-3: Legacy ordinal markers get the same normalization: a spec supplying `buildLegacyMarker` probes with an undelimited query and still reports already-filed when the stub matches.
- [ ] AC-4: A duplicate inside the search-index window is caught: with the search probe stubbed to return empty and the strongly-consistent issue-list stub returning a body containing the marker, `graduate()` records the finding as skipped `already-filed` and `createFollowUpIssue` is never spawned.
- [ ] AC-5: The strong read runs only on the would-file path: when the search probe already reports a match, no `gh issue list` spawn occurs.
- [ ] AC-6: Two source buckets in one `graduateRetroProposals` invocation cannot file the same marker twice — given identical categories in the `framework` and `consumer` buckets and both probes stubbed empty, exactly one `gh issue create` spawn occurs.
- [ ] AC-7: An undecidable probe still files rather than swallowing: with every probe stub erroring, the finding is filed and appears in `filed[]`.
- [ ] AC-8: The doc comments no longer assert the disproven claim: `grep -rn "indexable via .gh search" .agents/scripts/lib/feedback-loop/` returns no match.

## Verify
- node --test tests/lib/feedback-loop/graduator-core.test.js (unit)
- node --test tests/lib/retro-proposals-graduator.test.js (unit)
- npm test (validate)
- npm run lint (validate)

## References
- `.agents/scripts/lib/orchestration/run-epilogue.js` — exists
- `.agents/scripts/lib/orchestration/story-follow-ups.js` — exists

## Non-Goals
- Guarding `run-epilogue.js` against double-invocation of the follow-up rollup. A correct dedup makes a repeated rollup harmless, and there is no in-repo evidence of the double-invocation mechanism to fix — chasing it would be speculative.
- Changing the marker format, the `contentFingerprint` derivation, or the `(epicId, category)` identity the retro marker encodes — the identity is correct; only the probe that reads it is broken.
- Changing the degrade-toward-filing posture on an undecidable probe.
- Retroactively closing or merging the consumer duplicates already filed in error (Beestera/swarm-os#655, #662, #664).

dsj1984:
Closes #4657

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes dedup behavior for all graduator follow-up filing (search query shape plus an extra gh list before create); mistakes could cause missed skips or extra API load, but failures still prefer filing over silent drops.
> 
> **Overview**
> Fixes feedback-loop graduators that **re-filed follow-ups on every rollup** because `gh search issues` was queried with the full HTML-comment marker (`<!-- … -->`), which GitHub’s index does not match; probes now search the **delimiter-stripped** marker text while bodies still embed the wrapped comment.
> 
> On the **would-file** path only, adds a **label-scoped `gh issue list --state all`** check so duplicates filed seconds earlier in the same run are caught when search is still eventually consistent. Probe failures still **degrade toward filing** (duplicate risk over dropping findings).
> 
> **Retro proposals** pass a **shared in-process `filedMarkers` set** across framework and consumer buckets so the same category cannot create two issues in one invocation when markers collide. Tests cover normalization, strong read, memo dedup, and legacy marker probing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 622cc2e1c72e096589f70dd57a27a7c40ba219da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/feedback-loop/audit-results-graduator.js
+++ b/.agents/scripts/lib/feedback-loop/audit-results-graduator.js
@@ -97,7 +97,9 @@ function metaSourceLabel(source) {
  * ({@link buildContentMarker}) at the Story #4415 cutover, but still
  * probed for so follow-ups filed before the cutover are recognized and
  * not re-filed. An HTML comment so it survives markdown rendering without
- * leaking into the visible body, but stays indexable via `gh search`.
+ * leaking into the visible body; the idempotency probe strips the comment
+ * delimiters before querying `gh search` (the raw `<!-- … -->` form never
+ * matches the index — Story #4657).
  *
  * @param {number} epicId
  * @param {number} index — zero-based finding ordinal within the Epic.
@@ -112,7 +114,10 @@ export function buildIdempotencyMarker(epicId, index) {
  * follow-up bodies. Derived from the finding's `lens|path|summary` triple
  * so the marker is stable across sibling insert/remove/reorder churn in
  * the source `audit-results` comment (Story #4415). An HTML comment so it
- * survives markdown rendering but stays indexable via `gh search`.
+ * survives markdown rendering without leaking into the visible body; the
+ * idempotency probe strips the comment delimiters before querying
+ * `gh search` (the raw `<!-- … -->` form never matches the index —
+ * Story #4657).
  *
  * @param {number} epicId
  * @param {{ lens?: string, path?: string, summary?: string }} finding
--- a/.agents/scripts/lib/feedback-loop/graduator-core.js
+++ b/.agents/scripts/lib/feedback-loop/graduator-core.js
@@ -288,7 +288,7 @@ export async function probePathStatus({
  * @param {string} marker
  * @returns {string}
  */
-export function normalizeMarkerQuery(marker) {
+function normalizeMarkerQuery(marker) {
   if (typeof marker !== 'string') return '';
   return marker.replaceAll('<!--', '').replaceAll('-->', '').trim();
 }
@@ -361,7 +361,7 @@ export async function probeMarkerExists({
  * @param {number} [opts.timeoutMs]
  * @returns {Promise<boolean>}
  */
-export async function confirmMarkerFiled({
+async function confirmMarkerFiled({
   marker,
   owner,
   repo,
--- a/.agents/scripts/lib/feedback-loop/graduator-core.js
+++ b/.agents/scripts/lib/feedback-loop/graduator-core.js
@@ -274,11 +274,33 @@ export async function probePathStatus({
 }
 
 /**
+ * Normalize an idempotency marker into a `gh search issues` query. Markers
+ * are HTML comments (`<!-- … -->`) so they survive markdown rendering
+ * without leaking into the visible body, but the `<` / `>` delimiters are
+ * NOT index-safe as a query: GitHub full-text search DOES index the text
+ * inside an HTML comment, yet a query that carries the `<!--` / `-->`
+ * delimiters never matches that indexed text (measured against this repo,
+ * Story #4657). Stripping the delimiters and trimming yields the bare marker
+ * text — `retro-proposal-followup: epic-1-<fp>` — which the index matches.
+ * The caller-facing marker is left untouched; normalization is the probe's
+ * own concern.
+ *
+ * @param {string} marker
+ * @returns {string}
+ */
+export function normalizeMarkerQuery(marker) {
+  if (typeof marker !== 'string') return '';
+  return marker.replaceAll('<!--', '').replaceAll('-->', '').trim();
+}
+
+/**
  * Probe whether a follow-up issue carrying the given idempotency marker
  * already exists in the routed repo. Uses `gh search issues` so we hit
- * the body field directly. Returns `true` when at least one match is
- * present; degrades to `false` on any spawn/parse error (better to risk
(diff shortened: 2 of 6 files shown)


dsj1984:
## 🔬 Code Review — Story #4657

**Comparison**: `main` … `story-4657`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

Issue #4721, opened 2026-07-23
doctor: commands-in-sync ignores the .agents/local/workflows/ projection source — permanent '1 stale' false positive


dsj1984:
## Symptom

In a consumer with a local workflow (e.g. mandrel-bench's `.agents/local/workflows/benchmark.md`), `mandrel doctor` on v2.10.0 reports:

```
✘  commands-in-sync       1 stale
   → Run `npm run sync:commands` to regenerate the `.claude/commands/` tree.
```

but the remedy is a no-op (`✔ 0 file(s) synced, 24 total commands`) — the failure can never clear, which also fails the `mandrel update` doctor phase on every upgrade.

## Cause

`runCommandsInSync` (lib/cli/registry.js) builds its expected set from `.agents/workflows/*.md` only (with the #4482 `command: false` opt-out honored). Since 1.75.0 (#4244) `sync-claude-commands.js` also projects `.agents/local/workflows/*.md` as a second, prune-exempt source, so a consumer-authored local command appears in `.claude/commands/` with no counterpart in the dir doctor scans → counted as `extra` → "1 stale".

The check's own docstring says it should "dry-run the sync-claude-commands logic" — it drifted from the script when the local projection source was added.

## Fix

Include `.agents/local/workflows/*.md` (same top-level-only + `command: false` filtering) in the check's source set, mirroring the sync script. A regression test: a dest command backed only by a local workflow must not count as stale.

Observed on mandrel v2.10.0, consumer dsj1984/mandrel-bench (doctor otherwise green, 11/12).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dsj1984:
**Superseded by #4724** — *doctor: include .agents/local/workflows/ in the commands-in-sync expected set* (`type::story`, `agent::ready`).

Planned via `/plan --tickets 4721`.

The analysis in this issue is preserved as the historical record; #4724 carries the delivery contract.

PR #4726, opened 2026-07-23
fix: doctor: include .agents/local/workflows/ in the commands-in-sync expected set (#4724)

### LINKED ISSUE #4724: doctor: include .agents/local/workflows/ in the commands-in-sync expected set
## Goal
Stop the permanent '1 stale' commands-in-sync false positive in consumers with local workflows by teaching the doctor check the second projection source the sync script has had since 1.75.0 (#4244).

## Spec
`runCommandsInSync` (lib/cli/registry.js) must build its expected command set as the union of BOTH projection sources — `.agents/workflows/` and `.agents/local/workflows/` — mirroring `sync-claude-commands.js` exactly, because the check's contract is to dry-run that script's logic.

Invariants (each mirrors an observable sync-script behavior):

- **Per-source filtering parity.** Each source dir gets the same filtering the payload dir gets today: dot-prefixed files skipped, frontmatter `command: false` excluded (#4482), unreadable source treated as projected.
- **Union semantics on basename collision.** The sync script's payload-wins shadowing (`byRel`) means a basename projects iff at least one source's copy is non-excluded — including the payload-excluded/local-projected case, where the excluded payload entry does NOT enter `byRel` and so does not shadow the local copy. A per-dir filter-then-union of basenames reproduces this exactly; do not model shadowing as payload-side-only exclusion.
- **Absent local dir degrades silently.** The existing `readDir` seam already returns `[]` on an unreadable dir; a consumer with no `.agents/local/workflows/` must behave byte-identically to today.
- **loops/ stays out of scope on both sides.** The check compares top-level basenames only and its dest listing never sees `loops/*.md`; keep that symmetry.

Test-contract ripple: two existing tests in tests/cli/registry.test.js encode the old two-call `readDir` order (callCount-based src/dest fakes, and the #3588 `dirsRead` deepEqual). They must be updated for the third dir read — prefer dir-keyed fakes (the `registry-agents-in-sync.test.js` style) over call counting so the order stops being load-bearing.

## Changes
- `lib/cli/registry.js` — refactors-existing
- `tests/cli/registry.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: A destination command backed only by a `.agents/local/workflows/*.md` source is counted as in-sync — the check no longer reports it as 'stale' (regression for the mandrel-bench benchmark.md case).
- [ ] AC-2: A local workflow with no projected destination counterpart is reported as 'not synced' (local sources gain missing-detection, not just stale-suppression).
- [ ] AC-3: A local source with frontmatter `command: false` is excluded from the expected set, and a basename supplied by both dirs counts once — including the payload-excluded/local-projected shadowing case.
- [ ] AC-4: A consumer with no `.agents/local/workflows/` directory behaves exactly as before (all pre-existing commands-in-sync tests still pass, updated only for the added dir read).
- [ ] AC-5: The check's docstring names both projection sources so it matches the sync script it dry-runs.

## Verify
- node --test tests/cli/registry.test.js (unit)
- npm test (validate)

## Non-Goals
- No change to sync-claude-commands.js — the script is the reference behavior, not the defect.
- No loops/ namespace coverage in the check — dest listing is top-level-only today and stays so.
- No doctor-surface or remedy-text changes beyond the corrected expected set.

dsj1984:
Closes #4724

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Low Risk**
> Scoped to doctor parity logic and unit tests; no auth, data, or runtime delivery paths change.
> 
> **Overview**
> Fixes **`mandrel doctor`** falsely reporting **stale** `.claude/commands/` entries when a command exists only because **`sync-claude-commands`** projected a workflow from **`.agents/local/workflows/`** (consumer-authored source since 1.75.0 / #4244).
> 
> The **`commands-in-sync`** check now builds its expected command set from the **union** of **`.agents/workflows/`** and **`.agents/local/workflows/`**, still honoring **`command: false`** opt-outs and matching sync’s **payload-wins** shadowing (a payload-excluded basename does not block a projectable local copy). Missing local dir continues to behave like an empty source.
> 
> Tests gain a path-keyed **`readDir`** helper and cover local-only dest parity, missing local projections, local exclusions, and basename collision cases (**#4721**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40a58626679780c9254a3b641865eda69899b140. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4724

**Comparison**: `main` … `story-4724`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4770, opened 2026-07-25
fix: stop close-validation gate capture from stalling the child's stdout pipe (#4766)

### LINKED ISSUE #4766: Stop close-validation gate capture from stalling the child's stdout pipe
## Goal
A high-volume gate child must never be killed by the close-validation output plumbing, so a green lint verdict cannot present as a failed close.

## Spec
**Mechanism (measured, sharper than the source report).** `pipePrefixed` in `lib/close-validation/process.js` calls `emit()` **synchronously** inside the child stream's `'data'` handler, and on the capture path `emit` lands on `fs.writeSync` in `single-story-close/gate-log.js`. One synchronous file write per line blocks the event loop while the child keeps writing, the OS pipe buffer fills, and the child's non-blocking write returns `EAGAIN`. Biome's `biome_console` `.unwrap()`s that error and aborts with exit 101 — a panic, not a lint violation. It is volume-dependent and `npx biome ci .` already emits ~625 lines on a clean `main`, so any close can hit it.

**Invariant to establish.** Draining the child's stdout/stderr must not perform synchronous or unbounded per-line work. The child must never observe a stalled reader on a non-blocking pipe, regardless of output volume.

**Load-bearing constraint on the fix.** Per-line `[gate-name] ` prefixing exists so concurrent gates do not interleave unreadably. Handing the child a raw file descriptor removes the JS reader entirely but also removes prefixing, so that route is only acceptable if per-gate attribution is preserved another way (a per-gate artifact, or a documented single-gate-per-file guarantee). Mechanism is otherwise the deliverer's choice.

**Explicitly not the remedy.** A re-run is forbidden by the no-flaky-rerun rule; this is a race that always lands on the first gate of every close. `AGENT_LOG_LEVEL=verbose` is **not** a workaround — it fixes lint by restoring inline streaming but then fails `coverage-capture`, because two tests assert default-level behaviour and the stdout-budget assertion measured 308889B against a ~2KB ceiling.

**Failure-path behaviour is unchanged.** `replay()` of the captured tail on gate failure, the success-path one-line `digest()`, and gate exit-code propagation must all survive.

## Changes
- `.agents/scripts/lib/close-validation/process.js` — refactors-existing
- `.agents/scripts/lib/orchestration/single-story-close/gate-log.js` — refactors-existing
- `tests/close-validation-gate-output-volume.test.js` — creates

## Acceptance
- [ ] AC-1: A gate child emitting output volume at least an order of magnitude above the ~625-line clean-`main` Biome baseline completes through the capture sink with a zero exit status and no lost lines.
- [ ] AC-2: Draining the child's stdout/stderr performs no synchronous per-line file write: the `'data'` path no longer calls `fs.writeSync` per line, directly or transitively.
- [ ] AC-3: Every line the child emitted is present in the gate log artifact, including a trailing line with no terminating newline.
- [ ] AC-4: Per-gate attribution survives: concurrent gates remain distinguishable in the captured output, either by retained `[gate-name] ` prefixing or by a per-gate artifact.
- [ ] AC-5: Gate exit codes still propagate unchanged, and a non-zero gate still fails close-validation with the same error text and hint.
- [ ] AC-6: The failure path still replays the captured tail inline, and the success path still emits the one-line digest naming the artifact.
- [ ] AC-7: The verbose inline-streaming path (`AGENT_LOG_LEVEL=verbose`) is unchanged, and the default-level stdout budget asserted by the coverage-capture tests continues to hold.
- [ ] AC-8: A regression test drives the high-volume case through the real sink and would fail against the pre-fix synchronous-write path.

## Verify
- node --test tests/close-validation-gate-output-volume.test.js (unit)
- node --test tests/lib/orchestration/single-story-close/gate-log.test.js (unit)
- node --test tests/lib/close-validation-gate-helpers.test.js (unit)
- npm test (contract)
- npm run lint (validate)

## Non-Goals
- Not fixing the 29 pre-existing Biome useOptionalChain / noUnusedVariables warnings — that cuts the trigger volume but leaves the race latent for any future noisy gate, and it is independently worth doing as its own change.
- Not passing --reporter=summary to biome ci — a Biome-specific mitigation that becomes redundant once the drain path is safe, and it would cost full diagnostics on the gate path.
- Not upgrading Biome to chase an upstream biome_console fix — a version bump is its own migration motion and would leave the class unfixed for every other gate.
- Not changing AGENT_LOG_LEVEL semantics or the verbose inline-streaming path.

dsj1984:
Closes #4766

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Touches core close-validation I/O plumbing used on every story close; behavior change is intentional (async artifact, close vs exit) but well-tested with volume regressions.
> 
> **Overview**
> Fixes close-validation failing when chatty gates (e.g. Biome) emit hundreds of lines: **synchronous per-line artifact writes** in the gate log sink blocked the event loop, filled the child's stdout pipe, and could kill an otherwise green gate (e.g. Biome exit 101).
> 
> **Gate log sink** now buffers lines through an async `fs.WriteStream` instead of `fs.writeSync`, and exposes **`flush()`** so callers await a complete artifact before read/replay. **Close-validation phase** wraps gate runs in `try/finally` and always `await gateLog.flush()`.
> 
> **Gate runner (`process.js`)** resolves on child **`close`** (not `exit`) so status is reported only after stdio is drained; **`pipePrefixed`** uses an O(chunk) line scan and swallows stream errors; the Biome "no files processed" probe keeps a **32-line tail** instead of unbounded string growth.
> 
> Adds **volume regression tests** (50k lines, no `writeSync` on drain path; high-volume spawn; concurrent gate prefixes) and updates existing gate-log tests for async flush.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f50ed9cb50dba79e1e422f2f06369eb589ef354b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/close-validation/process.js
+++ b/.agents/scripts/lib/close-validation/process.js
@@ -11,22 +11,48 @@ import { spawn } from 'node:child_process';
  * Pipe a child stream's output line-by-line through `emit`, prepending
  * `prefix` to each line. Tail bytes without a trailing newline flush on
  * `end` so the operator never loses the last line of a gate's output.
+ *
+ * ## The drain must stay cheap (Story #4766)
+ *
+ * This handler runs on the reader side of the child's stdout/stderr pipe.
+ * Every microsecond spent here is a microsecond the pipe is not being read,
+ * and once the OS pipe buffer fills, the child's own write blocks — or, on a
+ * non-blocking pipe, fails outright with `EAGAIN`. A gate child is not
+ * obliged to survive that: Biome's `biome_console` `.unwrap()`s the error and
+ * aborts the process with exit 101, so a green lint verdict presents as a
+ * failed close. Two consequences bind everything on this path:
+ *
+ *   1. Splitting is O(chunk), not O(chunk × lines) — the scan advances a
+ *      `start` index instead of re-slicing the buffer once per line, so a
+ *      64KB chunk carrying 500 lines does not copy 16MB.
+ *   2. **`emit` MUST NOT block.** A synchronous per-line file write is
+ *      exactly the stall this path cannot afford; the close path's capture
+ *      sink (`single-story-close/gate-log.js`) buffers to an async stream for
+ *      that reason.
  */
 function pipePrefixed(stream, prefix, emit) {
   let buf = '';
   stream.setEncoding('utf8');
   stream.on('data', (chunk) => {
     buf += chunk;
-    while (true) {
-      const nl = buf.indexOf('\n');
-      if (nl === -1) break;
-      emit(prefix + buf.slice(0, nl));
-      buf = buf.slice(nl + 1);
+    let start = 0;
+    let nl = buf.indexOf('\n', start);
+    while (nl !== -1) {
+      emit(prefix + buf.slice(start, nl));
+      start = nl + 1;
+      nl = buf.indexOf('\n', start);
     }
+    if (start > 0) buf = buf.slice(start);
   });
   stream.on('end', () => {
-    if (buf.length > 0) emit(prefix + buf);
+    if (buf.length > 0) {
+      emit(prefix + buf);
+      buf = '';
+    }
   });
+  // A pipe-level error (EIO on a vanished child) must not become an
+  // unhandled 'error' event that takes the whole close down.
+  stream.on('error', () => {});
 }
 
 /** Wire the AbortSignal so an abort kills the child. Returns the cleanup fn. */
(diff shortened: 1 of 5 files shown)


dsj1984:
## 🔬 Code Review — Story #4766

**Comparison**: `main` … `story-4766`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4777, opened 2026-07-26
fix(crap): join per-method coverage in original-source coordinates — TS scoring resolves ~5% of methods (#4775)

### LINKED ISSUE #4775: fix(crap): join per-method coverage in original-source coordinates — TS scoring resolves ~5% of methods
## Goal

Make the CRAP gate score every scorable method instead of ~5% of them, by joining escomplex methods to istanbul coverage in a single line-coordinate system at scoring time.

## Spec

**Root cause.** `transpileIfNeeded` (`.agents/scripts/lib/transpile.js:61`) transpiles TS/TSX via `ts.transpileModule` **without** `sourceMap`, so escomplex reports each method's `lineStart` in *transpiled* coordinates. `coverageForMethodInEntry` (`.agents/scripts/lib/coverage-utils.js:204-207`) then looks that line up against istanbul's `fnMap`, which is in *original source* coordinates, by exact equality:

```js
if (!idx.fnByStartLine.has(startLine)) return null;   // → coverage null → crap null → row dropped
```

The lookup returns `null`, `analyzeOnce` (`crap-utils.js:247-250`) sets `crap = null`, and the row is dropped at `crap-utils.js:441`. Silently.

**Evidence.** `apps/api/src/app.ts` in the swarm-os consumer — 297 source lines transpile to 195. The same 10 methods, two coordinate systems, zero overlap:

```
escomplex lineStart : 48, 50, 86, 95, 128, 132, 134, 145, 175, 192
istanbul fnMap      : 129, 131, 176, 184, 218, 225, 227, 239, 270, 293
```

Measured method-resolution rate across three all-TypeScript consumers, counting only files that **do** have a coverage entry:

| consumer | files w/ coverage | methods in them | resolved | rate |
|---|---|---|---|---|
| Beestera/swarm-os | 231 | 1618 | 100 | 6.2% |
| dsj1984/athportal | 511 | 3092 | 135 | 4.4% |
| dsj1984/domio | 341 | 1860 | 117 | 6.3% |

domio rules out coverage completeness as the cause: its artifact covers 341 of 374 files (only 51 methods sit in uncovered files) and it still resolves 6.3%.

**Coincidental matches are mis-attributions.** Because the two sets are in different coordinate systems, a "match" is a numeric collision that joins a method to an unrelated function's coverage. Consumer baselines are therefore unreliable, not merely sparse — every floor currently pinned to observed maxima across the three consumers is derived from bad data.

**The drift is known, and the existing mitigation is in the wrong phase.** `tests/crap-utils.tsx-coverage-keying.test.js` documents it explicitly: *"escomplex's per-method `lineStart` will not generally match the source line number recorded in the coverage entry. The per-method lookup absorbs the drift via `compareCrap`'s line-drift fallback."* It does not. `compareCrap` (`baselines/kinds/crap.js:254`) reconciles **current rows against baseline rows** at comparison time; at scoring time the row has already been dropped and never reaches it. That test then lays its fixture coverage entry at the transpiled line number escomplex reports — a configuration that cannot occur in a real `coverage-final.json` — so it passes while the production path is broken.

**Fix, part 1 — remap.** `transpileIfNeeded` emits `sourceMap: true` and returns the map alongside `outputText`; the CRAP path remaps each method's `lineStart` to original coordinates before the coverage lookup. Verified viable against `app.ts` using Node's built-in `SourceMap` from `node:module`, so this adds **no new runtime dependency**:

```
transpiled 48 → original 129   (fnMap 129) ✓
transpiled 50 → original 130   (fnMap 131)
transpiled 86 → original 175   (fnMap 176)
```

**Fix, part 2 — tolerant match.** Exact equality is still insufficient after remapping: remapped starts land within ±1 of the `fnMap` decl line, because escomplex's method start and istanbul's `decl.start.line` do not always agree on the token. Replace exact-line equality with containment — select the innermost `fnMap` entry whose `loc` range contains the remapped line, falling back to nearest decl within a bounded window. Both parts are required; neither alone fixes the join.

**Fix, part 3 — honest `requireCoverage: false`.** The flag currently only prevents whole-file skipping; each method is still dropped individually when its coverage is unresolved, regardless of the flag, which makes it a no-op for baseline population. Under `requireCoverage: false` a method with no coverage entry MUST score as 0% coverage (`crap = c² + c`, the correct CRAP treatment of untested code) rather than being dropped. `requireCoverage: true` keeps today's skip-and-count behaviour.

**Fix, part 4 — fail closed on a thin result.** The updater currently persists a 100-row baseline built from 5023 dropped methods and logs it as success; that is what let this rot undetected for five weeks across three repos. Add a minimum method-resolution rate below which the updater refuses to write, printing resolved/total and the worst offending files.

**Fix, part 5 — invalidate stale baselines.** Rows scored by the old join are not comparable to rows scored by the new one. `kernelVersion()` (`baselines/kinds/crap.js:50`) returns the installed `typhonjs-escomplex` version and will not change here, and the existing scorer-drift check keys on `escomplexVersion`, so neither trips. Introduce a scoring-semantics stamp so a baseline written by the old scorer fails closed with explicit re-baseline guidance instead of being silently compared across incompatible scoring.

The maintainability path is unaffected — MI is module-level and does not use the coverage join. MI output for the same tree must not move.

## Changes

- `.agents/scripts/lib/transpile.js` — refactors-existing
- `.agents/scripts/lib/coverage-utils.js` — refactors-existing
- `.agents/scripts/lib/crap-utils.js` — refactors-existing
- `.agents/scripts/lib/workers/crap-worker.js` — refactors-existing
- `.agents/scripts/lib/workers/combined-mi-crap-worker.js` — refactors-existing
- `.agents/scripts/update-crap-baseline.js` — refactors-existing
- `tests/crap-utils.tsx-coverage-keying.test.js` — refactors-existing (fixture must use real original-source coordinates)
- `tests/lib/crap-utils.sourcemap-join.test.js` — creates
- `.agents/docs/quality-gates.md` — refactors-existing

## Acceptance

- [ ] AC-1: `transpileIfNeeded` returns an original-source line mapping for TS/TSX inputs; JS inputs remain a passthrough with no mapping cost and no behaviour change
- [ ] AC-2: Per-method coverage resolves in original-source coordinates — a TS fixture whose coverage entry is keyed at real source lines resolves 100% of its methods
- [ ] AC-3: Containment matching resolves the ±1 decl offset; a fixture where the remapped start is one line off the `fnMap` decl line still resolves
- [ ] AC-4: `tests/crap-utils.tsx-coverage-keying.test.js` no longer lays its coverage entry at transpiled line numbers, and its docstring's claim that `compareCrap` absorbs scoring-time drift is removed
- [ ] AC-5: Under `requireCoverage: false` a method with no coverage entry scores as 0% coverage (`crap = c² + c`) and appears in the baseline; under `requireCoverage: true` it is skipped and counted as today
- [ ] AC-6: `update-crap-baseline.js` refuses to persist a baseline whose method-resolution rate is below the configured floor, exits non-zero, and prints resolved/total plus the worst unresolved files
- [ ] AC-7: A baseline carrying the old scoring-semantics stamp fails closed on check with explicit re-baseline guidance, rather than being compared against new-semantics scores
- [ ] AC-8: The serial scorer and the worker-pool scorer produce byte-identical rows through the new join — the existing combined-parity test covers the remap path
- [ ] AC-9: The maintainability baseline for an unchanged tree is byte-identical before and after this change
- [ ] AC-10: Regenerating full-scope against a real TypeScript consumer checkout resolves ≥ 90% of methods in files that have coverage, up from the 4-6% recorded above

## Verify

- `node --test tests/lib/crap-utils.sourcemap-join.test.js` (unit)
- `node --test tests/crap-utils.tsx-coverage-keying.test.js` (unit)
- `node --test tests/lib/crap-utils.combined-parity.test.js` (unit)
- `node --test tests/lib/coverage-utils.test.js` (unit)
- `node --test tests/baselines-byte-identical-js-only.test.js` (unit)
- `npm run test:quick` (unit)
- `npm run lint` (validate)
- `npm run format:check` (validate)

## Non-Goals

- Changing the CRAP formula itself (`crap-engine.js`) — the scoring math is correct; only the coverage join is wrong
- Re-deriving quality-gate floors in any consumer repo — floors pinned to today's mis-attributed maxima must be re-derived, but that is consumer-side work tracked separately
- The absent CRAP baseline-refresh projection and full-scope drift detection — filed as a follow-up Story that depends on this one
- Improving any consumer's actual test coverage; this Story changes what the gate can *see*, not what is tested

dsj1984:
Closes #4775

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Touches quality-gate scoring, baseline persistence, and compare semantics across close/CI; misconfiguration could block updates until re-baseline, but behavior is intentionally fail-closed rather than silent wrong scores.
> 
> **Overview**
> Fixes the **CRAP per-method coverage join** so TypeScript/TSX methods resolve against istanbul in **original-source line coordinates**, with guards so thin or incomparable baselines cannot slip through.
> 
> **Scoring path:** TS transpile can opt into `{ withLineMap: true }` (Node `SourceMap`); method `lineStart` is remapped before lookup. Coverage matching is **exact line → innermost containing function → nearest decl within ±1**. Shared helpers (`methodRowsFromReport`, `finalizeMethodRows`, `prepareSourceForScoring`) keep serial, pool, and worker paths aligned. With `requireCoverage: false`, unresolved methods now score as **0% covered** instead of being dropped.
> 
> **Updater & config:** `update-crap-baseline.js` refuses to persist when join health is below **`minMethodResolutionRate`** (default 0.75, skipped under 25 joinable methods), with worst-file telemetry. Dogfood **`crap.floors.*.max`** moves **30 → 125** to reflect an honest rollup after more methods become visible.
> 
> **Baseline semantics:** CRAP envelopes stamp **`scoringSemantics: coverage-join-v2`**. `check-baselines` **fails closed** on stamp mismatch; head-vs-base compare is **skipped** when base and head semantics differ (floors still apply).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8222382bae1c53aad225f239831a0ca83c7d375. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agentrc.json
+++ b/.agentrc.json
@@ -34,7 +34,7 @@
     "quality": {
       "gates": {
         "crap": {
-          "floors": { "*": { "max": 30 } },
+          "floors": { "*": { "max": 125 } },
           "targetDirs": [".agents/scripts", "bin", "lib"],
           "ignoreGlobs": [
             ".agents/scripts/lifecycle-emit.js",
--- a/.agents/docs/configuration.md
+++ b/.agents/docs/configuration.md
@@ -174,6 +174,7 @@ top-level keys are validation errors.
 | `quality.gates.crap.targetDirs` | No | `string[]` or `{ append?, prepend? }` | — | Directories whose JS sources the CRAP gate scores. Mandrel ships a `src/`-centric default; projects whose executable code lives elsewhere (e.g. this repo's `.agents/scripts/`) override here. The framework default is intentionally not auto-discovered, so an override is the explicit, auditable signal. |
 | `quality.gates.crap.newMethodCeiling` | No | `integer` | — | — |
 | `quality.gates.crap.requireCoverage` | No | `boolean` | — | — |
+| `quality.gates.crap.minMethodResolutionRate` | No | `number` | — | Fail-closed floor on the per-method coverage JOIN (Story #4775): the fraction of methods that must resolve a coverage entry, counted only over files that HAVE one, before `update-crap-baseline.js` will persist. A broken join is silent by construction — unresolved methods are simply absent — so the updater refuses rather than writing a thin baseline and logging it as success. Not enforced below 25 joinable methods, where a diff-scoped run's rate is noise. Default 0.75; a healthy repo resolves ~98%. |
 | `quality.gates.crap.friction` | No | `object` | — | Nested configuration block. |
 | `quality.gates.crap.friction.markerKey` | No | `string` | — | — |
 | `quality.gates.crap.refreshTag` | No | `string` | — | — |
--- a/.agents/docs/quality-gates.md
+++ b/.agents/docs/quality-gates.md
@@ -400,6 +400,15 @@ joinable methods, where a diff-scoped run's rate is noise. A healthy repo
 resolves ~98%; the 4–6% signature of a coordinate-system mismatch is far below
 the floor.
 
+**Re-derive your floors after adopting this.** A `crap.floors` ceiling pinned
+before the fix was pinned to a maximum computed over the minority of methods
+the join could see, so it is not a real ceiling — it is an artefact. This
+repository's own `*.max` moved from `30` to `125` on the first honest scan
+(2215 → 4058 visible methods), not because anything got worse but because the
+worst methods were previously invisible. `newMethodCeiling` is deliberately
+left at 30: the ratchet's forward pressure on *new* code is unaffected by how
+much old debt the gate can now see.
+
 **Old baselines are invalidated explicitly.** Rows scored by the previous join
 are not comparable to rows scored by this one, and neither `kernelVersion` nor
 `escomplexVersion` moves (both track the same upstream package). The envelope
--- a/.agents/docs/quality-gates.md
+++ b/.agents/docs/quality-gates.md
@@ -361,6 +361,53 @@ There is no CI guardrail rejecting unlabeled baseline edits; the convention is
 preserved so the operator can grep refresh commits in a PR diff, but
 self-policing is the operator's job during `/deliver`'s watch loop.
 
+### The per-method coverage join (Story #4775)
+
+CRAP is the only gate that joins two independently-produced artifacts: the
+per-method complexity escomplex derives from the source, and the per-function
+coverage istanbul derives from the test run. Everything below exists because
+that join is silent when it fails — an unresolved method is simply absent from
+the baseline, so a broken join looks exactly like a small repo.
+
+**One coordinate system.** For a TS/TSX source, escomplex parses the
+*transpiled* output and reports each method's `lineStart` in transpiled
+coordinates, while `coverage-final.json` is keyed against the *original*
+source. The scorer therefore asks `transpileIfNeeded` for a source map
+(`{ withLineMap: true }`, backed by Node's built-in `SourceMap` — no extra
+runtime dependency) and remaps each method start into original coordinates
+before the lookup. JavaScript is a passthrough: its coordinates already are
+original coordinates, so no map is computed and nothing changes. The
+maintainability path never requests a map, and the emitted code is
+byte-identical either way, so MI scores are unaffected.
+
+**Tolerant matching.** Remapping alone is insufficient: escomplex's method
+start and istanbul's `decl.start.line` disagree by a line when a decorator, a
+leading `export`, or a wrapped parameter list sits between them. The lookup is
+exact-line first (so every already-resolving row keeps its exact prior value),
+then innermost containment, then nearest declaration within ±1.
+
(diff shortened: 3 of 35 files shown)


dsj1984:
## 🔬 Code Review — Story #4775

**Comparison**: `main` … `story-4775`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4809, opened 2026-07-27
feat: route every suite temp dir under one reaped per-process root and guard the OS temp root against regression (#4808)

### LINKED ISSUE #4808: Route every suite temp dir under one reaped per-process root and guard the OS temp root against regression
## Goal
Mandrel's own suite leaks tens of thousands of `os.tmpdir()` directories per run into a temp root shared with self-hosted CI runners, where entry count — not size — drives a measured job-timeout failure mode; this Story makes teardown structural rather than per-call-site, and gives `check-test-temp-hygiene.js` the OS-temp-root dimension it never had.

## Spec
**Attribution before assertion.** A guard over a shared `os.tmpdir()` cannot attribute an entry to this suite, so every managed dir MUST be minted *inside* a single per-process suite root (`mandrel-suite-<pid>-<random>`), never as a direct sibling of unrelated processes' entries. That is what lets the assertion be exact (`no suite root survives`) with no prefix allowlist to rot and no false positive from a concurrent process.

**Creator-only reaping.** The scratch root is published to child processes via `MANDREL_TEST_TEMP_ROOT`, so a process that *inherited* an absolute root MUST NOT remove it — only the process that minted it may reap. The existing per-process memos are already the "I created it" signal; reaping MUST key off those, not off the env var's presence. Reaping MUST be registered such that it also runs when the suite fails, and MUST NOT convert a teardown error into a suite failure.

**The guard gains a dimension; it does not lose one.** The existing repo-`temp/` NDJSON snapshot/assert contract is unchanged and MUST keep passing. The new dimension rides the same `--snapshot` / `--assert` bracket CI already runs, so no CI wiring changes.

**Consumer safety.** `check-test-temp-hygiene.js` ships in the materialized `.agents/` payload. The runtime dimension is inherently safe for consumers (a suite that mints no suite root is vacuously green). The static dimension is not, so it MUST be scoped to this repo's own test globs and MUST NOT fire on a consumer's tests.

## Changes
- `.agents/scripts/lib/test-temp.js` — creates
- `.agents/scripts/lib/test-env.js` — refactors-existing
- `.agents/scripts/lib/config/temp-paths.js` — refactors-existing
- `.agents/scripts/check-test-temp-hygiene.js` — refactors-existing
- `tests/lib/test-temp.test.js` — creates
- `tests/check-test-temp-hygiene.test.js` — refactors-existing
- `tests/check-dead-exports.test.js` — refactors-existing
- `tests/lib/worktree/node-modules-strategy.test.js` — refactors-existing
- `tests/lib/test-env.test.js` — refactors-existing
- `docs/ci-contract.md` — refactors-existing

## Acceptance
- [ ] AC-1: A shared managed-temp module exports a helper that returns a fresh directory and registers its teardown, and every directory it returns is nested inside one per-process suite root named `mandrel-suite-<pid>-<random>`.
- [ ] AC-2: When the creating process exits — including after a failing test — the suite root and everything under it is removed; a teardown failure is reported but does not change the process exit code.
- [ ] AC-3: A process that inherited an absolute `MANDREL_TEST_TEMP_ROOT` from its parent does not remove that root on exit, while the process that minted it does.
- [ ] AC-4: The scratch roots minted by `.agents/scripts/lib/test-env.js` and `.agents/scripts/lib/config/temp-paths.js` are reaped by their creating process on exit.
- [ ] AC-5: All 25 repo test files that previously minted `os.tmpdir()` directories with no teardown obtain them from the shared helper instead; `git grep` for a `mkdtemp` call taking `os.tmpdir()` directly returns no hits under `tests/` or `.agents/**/__tests__/`.
- [ ] AC-6: `check-test-temp-hygiene.js --assert` exits non-zero and names the surviving paths when one or more `mandrel-suite-*` roots remain in the OS temp root, and exits zero when none do.
- [ ] AC-7: `check-test-temp-hygiene.js --assert` exits non-zero and names the offending file and line when a repo test file calls `mkdtemp` against `os.tmpdir()` directly rather than through the helper; the rule is scoped to this repo's test globs and does not fire on files outside them.
- [ ] AC-8: The pre-existing repo-`temp/` stream-pollution dimension of the guard is unchanged: its snapshot, assert, and `--clean` behaviours and their tests still pass.
- [ ] AC-9: Running the full suite leaves the OS temp root's entry count unchanged net of the run — measured before and after, counting only entries the suite is responsible for.
- [ ] AC-10: `docs/ci-contract.md` describes the guard's OS-temp-root dimension alongside the existing repo-`temp/` dimension.

## Verify
- node --test tests/lib/test-temp.test.js (unit)
- node --test tests/check-test-temp-hygiene.test.js (unit)
- node --test tests/lib/test-env.test.js (unit)
- Count `os.tmpdir()` entries, run `npm test`, re-count: the delta attributable to the suite is zero and no `mandrel-suite-*` root survives (e2e)
- node .agents/scripts/check-test-temp-hygiene.js --assert (validate)
- npm test (validate)
- npm run lint (validate)

## Non-Goals
- Touching the ~167 test files that already reap their own `mkdtemp` results — they are not leaking and rewriting them buys nothing.
- Changing the repo-`temp/` NDJSON stream-pollution dimension of the guard, its snapshot baseline location, or its `--clean` mode.
- Adding or reordering CI steps: the `--snapshot` / `--assert` bracket already exists and the new dimension rides it.
- Reaping temp entries this suite did not create, including the unrelated 57 GB gitleaks leak tracked separately.
- Any change to consumer-facing behaviour of `mandrel sync` or the materialized payload beyond the guard's new runtime dimension.

dsj1984:
Closes #4808

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Large, cross-cutting change to test infrastructure and CI gates rather than product runtime, but a mistake in reaping or hygiene assertions could cause flaky CI or mask real leaks on shared runners.
> 
> **Overview**
> **Stops OS temp root pollution** by replacing scattered `mkdtemp(os.tmpdir())` usage with **`makeTempDir()`** from a new **`.agents/scripts/lib/test-temp.js`**. Each process gets one **`mandrel-suite-*`** root; nested fixture dirs live under it and are **reaped on exit** so teardown is structural rather than per-call-site.
> 
> **Scratch seams** (`test-env.js`, `config/temp-paths.js`) now register **`reapOnExit`** for creator-minted `mandrel-test-temp-*` dirs (children that inherit `MANDREL_TEST_TEMP_ROOT` still do not reap the parent).
> 
> **`check-test-temp-hygiene.js`** gains two more assert dimensions alongside the existing repo **`temp/`** stream check: **surviving suite roots** in `os.tmpdir()` (snapshot diff vs pre-existing roots), and an **opt-in `--lint-globs`** scan for tests that call `mkdtemp` against `os.tmpdir()` without `makeTempDir` (with `test-temp-allow` escape). **CI** runs assert with **`--lint-globs`** after tests; **`docs/ci-contract.md`** documents the bracket.
> 
> Wide **test and `.agents` test migrations** to `makeTempDir`; hygiene and **`test-temp`** behavior covered by expanded tests; **dead-exports** baseline updated for new exports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f93fd630812726537f96f1a1fed52b10aa7c700. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4808

**Comparison**: `main` … `story-4808`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

#### 🟢 Lint check passed with 3 warning(s) _[lint]_

Scoped lint reported 3 warning(s) on the changed surface. Treat as suggestions.

Issue #4824, opened 2026-07-29
feedback-loop: classify runtime-emitted friction as framework-sourced and count recurrence across runs


dsj1984:
## Goal
Consumer projects capture framework friction correctly but never report it upstream, because every signal is tagged consumer and every recurrence is counted inside a single run.

## Spec
Two defects make the framework limb of the feedback loop unreachable. Both are in already-built machinery; neither needs new subsystems.

Classification contract. Today classifyPathSource requires positive evidence of a framework prefix in failingPath or command, and defaults to consumer. Runtime friction emitters populate neither field, so the default always wins. Invert the discriminator for runtime-emitted records: absence of an operator-supplied command or path, plus a framework-owned emitter.tool, classifies framework. Records that do carry an operator-supplied command or path keep the existing prefix scan unchanged, so a consumer command invoked through diagnose-friction stays consumer. Existing classifier behavior for both current inputs must not change — this adds a limb that is currently unreachable, it does not rewrite the ones in use.

Recurrence window. The threshold of two occurrences in retro-proposals is sound; its window is not. gatherRunFrictionSignals collects only the current run's Stories, so a defect firing exactly once per Story is discarded as a single occurrence on every Story. Widen the gather to reduce over the signal streams still present under the temp tree, keyed on category, emitter tool, and a details fingerprint. Auto-purge shortening that window is acceptable: a short window under-counts and therefore fails toward not filing, which is the safe direction.

Legibility. A roll-up that discards every candidate must name the recurring fingerprint and its cross-run count. Rendering it as no follow-ups is how this stayed invisible for eighteen consecutive Stories.

## Changes
- `.agents/scripts/lib/observability/source-classifier.js` — refactors-existing
- `.agents/scripts/lib/observability/signals-writer.js` — refactors-existing
- `.agents/scripts/lib/orchestration/story-follow-ups.js` — refactors-existing
- `.agents/scripts/lib/orchestration/retro-proposals.js` — refactors-existing
- `.agents/scripts/lib/orchestration/run-epilogue.js` — refactors-existing
- `tests/lib/observability/source-classifier.test.js` — refactors-existing
- `tests/lib/orchestration/story-follow-ups.test.js` — refactors-existing
- `tests/lib/orchestration/retro-proposals.test.js` — refactors-existing
- `tests/lib/orchestration/run-epilogue.test.js` — refactors-existing

## Acceptance
[...]

Issue #4821, opened 2026-07-29
Correct the always-on instruction spine and give its budget one honest definition


dsj1984:
## Goal
The always-on spine carries instructions that are wrong for an edit-based runtime, and the ratchet meant to bound it under-reports the real per-session tax by ~5KB, so neither the wrong instructions nor the growth they hide is currently visible.

## Spec
Current state (verified 2026-07-29): two ratchets claim to bound the always-loaded context and disagree by ~5KB. `check-context-budget.js` resolves the full transitive `CLAUDE.md` closure (6 files, 21,314 bytes, CI-gated at ci.yml) while `tests/bootstrap/always-on-closure.test.js` matches only `@.agents/**.md` (16,289 bytes against a 16,384-byte cap), so `AGENTS.md` and `.agentrc.json` growth is uncharged.

Invariants:

- **One definition of always-loaded.** The unit ratchet MUST derive its file set from `resolveAlwaysLoadedClosure()` in `.agents/scripts/lib/doc-tiers.js` — the resolver the CI gate already uses — never a private regex. Two disagreeing definitions is the defect; widening the regex would reproduce it in a new place.
- **Widening must not become cover for growth.** Charging more files raises the budget constant; it rises only to the measured post-trim total plus modest slack.
- **The security baseline is untouched.** The duplicated security bullet in § 3 of .agents/instructions.md is removed as duplication, not as a relaxation — .agents/rules/security-baseline.md stays the SSOT and keeps every sentinel MUST and section the existing AC-2 assertions pin, per § 1.K's inviolability carve-out.
- **Corrections outrank trims.** The § 4 whole-file mandate, the § 5 lettering gap, and the unconditional § 1.C docs MUST in .agents/instructions.md are why this Story exists; if byte pressure forces a choice, drop a trim, never a correction.
- **The § 5 renumbering is atomic with its two inbound citations** in .agents/rules/git-conventions.md and .agents/README.md, since check-doc-links.js runs inside npm run lint.

## Changes
- `.agents/instructions.md` — refactors-existing
- `tests/bootstrap/always-on-closure.test.js` — refactors-existing
- `.agents/rules/git-conventions.md` — refactors-existing
- `.agents/README.md` — refactors-existing
- `AGENTS.md` — refactors-existing

## Acceptance
[...]

Issue #4816, opened 2026-07-28
deliver: persist the close terminal envelope to disk and teach recovery a live-close state


dsj1984:
## Goal
Make the Story delivery return path survive a worker turn that ends before its close finishes, so an orphaned envelope costs a file read instead of a router recovery cycle plus a full worker resume.

## Spec
Current state (verified 2026-07-27): `emitTerminalEnvelope` (`lib/orchestration/story-deliver-terminal.js`) writes to stdout only — already the single writer behind every emit site. A worker turn that ends while its close still runs loses that stdout; `decideRecovery` then sees `agent::executing` + no PR, returns `executing-no-pr`, and asserts "Implementation never finished" — false during a live close, and its `single-story-init.js` next command can put two closes on one PR.

Contract:

1. **Persistence belongs to that single writer.** `emitTerminalEnvelope` also persists the same object to `<tempRoot>/orchestration/story-deliver-terminal-<storyId>.json` via a new helper beside `orchestrationLogDir()`, honouring `project.paths.tempRoot` and main-checkout anchoring — a close inside a Story worktree must write where the router reads. Atomic (write-then-rename) and best-effort: a failed persist must not alter the stdout envelope or the exit code, and must not throw. A null `storyId` writes nothing.
2. **Recovery reads artifacts, not guesses.** `decideRecovery` stays pure; a `closeArtifacts` probe (persisted envelope + `close-gates-<id>.log` mtime) joins its inputs. Ahead of `executing-no-pr`: a persisted envelope answers with its own status and `nextCommand`; otherwise a gate log inside the freshness window yields `close-in-flight`, whose detail warns that re-init races the live close. It joins `TRANSIENT_SHAPES`; `executing-no-pr` survives unchanged when neither artifact exists.
3. **The artifact must be reapable** — `temp-retention.js` scans `*.log` only today.
4. **The rule binds where the behaviour is decided.** `story-worker.md` — below the role-delta marker, inside the 8192-byte per-agent ceiling — forbids returning before the envelope and forbids backgrounding or delegating the close; `helpers/deliver-story.md` sends an envelope-less child turn to the file before any probe.

## Changes
- `.agents/scripts/lib/config/temp-paths.js` — refactors-existing
- `.agents/scripts/lib/orchestration/story-deliver-terminal.js` — refactors-existing
- `.agents/scripts/lib/orchestration/deliver-recover.js` — refactors-existing
- `.agents/scripts/lib/temp-retention.js` — refactors-existing
[...]

Issue #4815, opened 2026-07-28
deliver-light: make the gate's `proceed-light` answer representable, attended-only, and auditable


dsj1984:
## Goal
The light suitability gate offers the operator `escalate-plan` or `proceed-light`, but only the first is executable — no flag records a `proceed-light` answer, so a deliberately-coarse prediction heuristic becomes an unappealable veto whose only workaround is under-declaring the footprint.

## Spec
An attended-only operator override that satisfies the shape half of the light gate's conjunction, refusing wherever the objection is a risk rule rather than a size prediction.

**Objection codes.** `deriveStoryShape` gains a `code` on its decision (`null` on `lite`) — one stable identifier per rejection, covering the four ceiling rules, the two absolute rules (migration span, sensitive path), and the pre-rule rejections (undeclared/unreadable/glob footprint, empty acceptance, unavailable classification). Additive only: `route`, `reasons`, `shape`, `ceilings`, `preserves` unchanged, so plan-persist's backstop and `/deliver`'s dispatch read what they read today. Classification MUST key off the code, never the prose reason.

**Allowlist, not denylist.** Only the four ceiling codes are overridable, exported frozen from `light-suitability.js`; an unrecognized or future code is non-negotiable by construction.

**Three required conditions.** The override applies only when (1) the outcome would otherwise be `ask-operator` — inert on a gate that never objected, and it MUST NOT alter `deriveLightSuitability`'s honest `suitable` verdict; (2) the code is in the allowlist; (3) the ledgered verdict is already `lite` with a recorded reason, so an unaudited "trust me, it's small" still buys nothing. A refused override is recorded as a reason, never dropped.

**Attended-only.** With `--yes` it is a CLI usage error (exit 1); independently the pure core still resolves `escalate-plan` under `yes: true`, keeping "unattended over-scope fails closed" true of the core and not just the shell.

**Auditable.** Gate envelope and receipt Story body both record the overridden code, the gate's reason, and the operator's reason. No new label, no `.agentrc` key.

The diff backstop already bounds actual size against ground truth, so overriding a *prediction* weakens no real guard. Keep the classification in a small pure helper — both touched scripts sit under the maintainability ratchet.

## Changes
- `.agents/scripts/lib/orchestration/complexity-gate.js` — refactors-existing
- `.agents/scripts/lib/orchestration/light-suitability.js` — refactors-existing
- `.agents/scripts/deliver-light.js` — refactors-existing
[...]

PR #4840, opened 2026-07-29
fix: review: stop the scoped-lint gate failing open silently, and fix why its runner cannot execute (#4839)

### LINKED ISSUE #4839: review: stop the scoped-lint gate failing open silently, and fix why its runner cannot execute
## Goal
The scoped-lint review gate has effectively never run in the primary consumer, because a runner that cannot execute produces no finding and the review reports success anyway.

## Spec
Two problems, and the second is why the first went unnoticed for months.

Why the runner cannot execute. In review-providers/native.js, an `executionFailed` lint summary emits friction telemetry and no finding. The reason text is invariant across every occurrence: lint runner produced no parseable output, binary missing, parse failure, or environment issue. Measured 2026-07-29 it fired on 18 of the 23 Beestera/swarm-os Stories carrying friction — 78 percent — and on 5 Stories here. A failure that reproduces at that rate against two different repositories is not a flaky environment; something about how the scoped runner is invoked or how its output is parsed fails deterministically. Diagnose that first and fix the cause. If the cause turns out to be genuinely environmental, say so explicitly with the evidence rather than treating the symptom.

Why nobody noticed. Story #4699 deliberately made an unexecutable tool a degradation rather than a code finding, so severity counts reflect code findings only. That intent is correct and must be preserved — do not reclassify it as a finding to make it visible. The defect is that the review's own verdict cannot distinguish a lint gate that ran and found nothing from one that never ran. A degraded gate must be visibly degraded in the review outcome the operator and the close pipeline see. Whether that blocks the close or merely reports is a judgement call for the deliverer; what must not survive is a review reporting clean when a gate did not execute.

## Changes
- `.agents/scripts/lib/orchestration/review-providers/native.js` — refactors-existing
- `tests/lib/orchestration/review-providers/native.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: AC-1: The cause of the execution failure is identified and fixed, with the diagnosis recorded in the PR body rather than only in code: `node --test tests/lib/orchestration/review-providers/native.test.js` exits 0.
- [ ] AC-2: AC-2: A lint runner that cannot execute no longer yields a review outcome indistinguishable from a clean one — the degraded state is represented in the verdict the close pipeline reads: `node --test tests/lib/orchestration/review-providers/native.test.js` exits 0.
- [ ] AC-3: AC-3: Severity counts still reflect code findings only, so #4699's intent survives — an execution failure does not appear as a critical, high, or medium finding: `node --test tests/lib/orchestration/review-providers/native.test.js` exits 0.
- [ ] AC-4: AC-4: The friction telemetry emission is unchanged — an execution failure still emits its tool-degraded signal: `grep -q 'TOOL_DEGRADED' .agents/scripts/lib/orchestration/review-providers/native.js` exits 0.
- [ ] AC-5: AC-5: A lint runner that executes normally still produces its findings and a clean review, so the fix does not make every review look degraded: `node --test tests/lib/orchestration/review-providers/native.test.js` exits 0.
- [ ] AC-6: AC-6: Repo gates stay green with no baseline churn: `npm run lint`, `npm test`, and `node .agents/scripts/check-baselines.js` each exit 0.
- [ ] AC-7: AC-7: The CI-only ratchets stay green: `node .agents/scripts/check-dead-exports.js --production` and `node .agents/scripts/check-context-budget.js` each exit 0.

## Verify
- node --test tests/lib/orchestration/review-providers/native.test.js (unit)
- npm run lint (validate)
- npm test (unit)
- node .agents/scripts/check-dead-exports.js --production (validate)

## Non-Goals
- Reclassifying an unexecutable tool as a code finding. Story #4699 routed it to telemetry so severity tiers reflect code findings only; that intent stands and visibility must be achieved without inflating severity counts.
- Changing the friction telemetry emission itself, the tool-degraded category, or anything in the feedback loop delivered by #4824, #4828 and #4837 — this Story is the defect that loop detected, not a change to the loop.
- Correcting the source classification on issue 4836. It reads consumer because dominantSource took a corpus majority of rows stamped before #4824; it self-corrects as the corpus turns over.

dsj1984:
Closes #4839

## Root-cause diagnosis

The scoped-lint review gate reported `executionFailed` — and therefore emitted
**zero findings while the review rendered a clean verdict** — on 18 of 23
Beestera/swarm-os Stories carrying friction (78%) and on 5 mandrel Stories
(#4801, #4815, #4816, #4821, plus one during #4824's own close). The gate had
effectively never run in the primary consumer.

The cause was **neither environmental nor a parse failure.** It was reproduced
directly against this repository, and it is three separate deterministic
invocation defects that all collapse into the same symptom.

### 1. The markdown runner was never resolvable (the dominant cause)

The provider spawned `npx --no markdownlint`. The binary this project — and the
consumer — installs is **`markdownlint-cli2`**; there is no bare `markdownlint`
bin in `node_modules/.bin`. Measured:

```
$ npx --no markdownlint README.md --ignore node_modules
npm error could not determine executable to run     # exit 1, no other output
```

`parseLintOutput` was *already* written for cli2's `Summary: N error(s)` line, so
the invocation and the parser had never agreed with each other. The
`--ignore node_modules` flag was likewise `markdownlint-cli` v1 syntax that cli2
rejects.

Because `npx` exits non-zero and prints nothing a reporter regex can match, the
summary came back `parsed: false, status: 1` → `executionFailed: true`.

### 2. One runner's failure became the other's verdict

`runScopedLint` ran biome and markdownlint and then folded both into a **single**
`parseLintOutput` call over concatenated output and the maximum exit status. So
defect 1's exit 1 became biome's verdict too: **any** change set containing at
least one `.md` file degraded the whole gate whenever biome had nothing to
report — which is precisely the clean case the gate exists to confirm. That is
the 78%: nearly every Story diff touches a markdown file.

### 3. Biome's empty-scope exit was read as a broken runner

`biome lint` exits 1 with `No files were processed in the specified paths.` when
every supplied path is excluded by `biome.json` (`temp/`, `dist/`,
`.worktrees/`, anything in the VCS ignore file). Same shape: non-zero exit,
nothing parseable, `executionFailed: true`. That is an empty scope, not a runner
that could not execute.

### Reproduction, before and after

`runScopedLint` against real change sets in this repo:

| change set | before | after |
| --- | --- | --- |
| code only, clean | `executionFailed=false` | `executionFailed=false` |
| markdown only | **`executionFailed=true`** | `executionFailed=false` |
| mixed code + markdown | **`executionFailed=true`** | `executionFailed=false` |
| all paths biome-excluded | **`executionFailed=true`** | `executionFailed=false` |

And the gate now *detects markdown errors at all* — a `.md` file with four
violations previously produced `executionFailed` plus zero findings, and now
produces `{ errors: 4, parsed: true }` and one `high` finding.

## The second defect: why nobody noticed

Story #4699 correctly routed an unexecutable tool to friction telemetry rather
than making it a code finding, so severity tiers keep reflecting code findings
only. What it did not add was any *other* channel — so the review's own verdict
became unable to distinguish **"lint ran and found nothing"** from **"lint never
ran"**. Both rendered `✅ No findings` with an all-zero tally, and the close
pipeline read the second as the first.

A degradation now travels **beside** the `Finding[]`, never inside it
(`review-providers/degraded-gates.js`):

- the provider exposes `getDegradations()`; the chain merges across inline entries;
- `runCodeReview`'s envelope — the object the close pipeline reads — carries
  `degraded` and `degradations[]`;
- the structured comment gains a `### ⚠️ Degraded Gates (N)` section and **can no
  longer print `✅ No findings`** when a gate did not run;
- the close phase states it on both operator surfaces (progress stream and the
  Story cross-reference comment), and states it *always* — rendering `none` when
  healthy, so an absent line can never be read as a clean gate.

**#4699's intent is preserved exactly.** An execution failure still produces no
`Finding` at any tier (not even `suggestion`), `countBySeverity` is untouched,
and the `tool-degraded` friction emission is byte-for-byte unchanged.

### Judgement call: report loudly, do not block

A degraded gate does not halt the close. The close pipeline already runs the
canonical `npm run lint` as a **hard close-validation gate before** the review
phase; the review's scoped lint is a second, narrower read of the same surface.
Failing a merge because a *secondary* read of an already-gated surface could not
start would cost delivery without buying coverage. What was broken was the
silence, so the fix makes the silence impossible. Escalating to a block is a
one-line change in `degraded-gates.js` if the operator posture ever needs it.

## Notes

- Runner resolution, per-surface classification and the merge were extracted to
  `review-providers/scoped-lint.js`; the per-file MI tolerance (absolute 0.5)
  drove the extraction rather than a baseline refresh.
- No baseline file was rewritten. The production dead-exports ratchet reports
  `added=0 removed=51 (ok)`, matching main's own verdict.
- This review gate runs from the **main checkout's** code, so this PR's own
  review comment was produced by the pre-fix path and shows no degraded-gate
  line. That is expected, and it is the last time it can happen.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.agents/scripts/lib/orchestration/code-review.js
+++ b/.agents/scripts/lib/orchestration/code-review.js
@@ -44,6 +44,10 @@ import { resolveConfig } from '../config-resolver.js';
 import { computeChangeSet } from './change-set.js';
 import { deriveChangeLevel, resolveDepth } from './review-depth.js';
 import {
+  collectProviderDegradations,
+  degradationEnvelope,
+} from './review-providers/degraded-gates.js';
+import {
   countBySeverity,
   renderFindings,
 } from './review-providers/findings-renderer.js';
@@ -176,6 +180,7 @@ function resolveScopeEnvelope(opts, config) {
  *   postedCommentId: number|null,
  *   commentTargetId: number,
  *   halted: boolean,
+ *   degraded: boolean, degradations: Array<object>,
  *   blockerReason: string|null,
  * }>}
  */
@@ -357,6 +362,11 @@ async function executeReviewPipeline({ opts, config, envelope }) {
     logger,
   );
 
+  // Story #4839 — degraded gates ride beside the findings, never inside them.
+  const degradations = await collectProviderDegradations(
+    reviewProvider,
+    logger,
+  );
   const severity = countBySeverity(findings);
   const halted = hasSurvivingCritical(severity);
   const report = renderFindingsFn({
@@ -367,6 +377,7 @@ async function executeReviewPipeline({ opts, config, envelope }) {
     findings,
     provider: providerName,
     promptMessages,
+    degradations,
   });
 
   const { posted, postedCommentId } = await postReviewComment({
@@ -385,6 +396,7 @@ async function executeReviewPipeline({ opts, config, envelope }) {
     postedCommentId,
     commentTargetId,
     halted,
+    ...degradationEnvelope(degradations),
     blockerReason: halted
       ? `code-review reported ${severity.critical} critical blocker(s)`
       : null,
--- a/.agents/scripts/lib/orchestration/review-providers/degraded-gates.js
+++ b/.agents/scripts/lib/orchestration/review-providers/degraded-gates.js
@@ -1,0 +1,222 @@
+/**
+ * review-providers/degraded-gates.js — the review's "a gate did not run"
+ * channel (Story #4839).
+ *
+ * ## The defect this closes
+ *
+ * Story #4699 correctly decided that a tool which **could not execute** is an
+ * operational degradation, not a code finding: it is routed to friction
+ * telemetry so severity tiers keep reflecting code findings only. What #4699
+ * did not add was any *other* channel, so the review's own verdict became
+ * unable to distinguish "lint ran and found nothing" from "lint never ran" —
+ * both rendered `✅ No findings` with an all-zero severity tally, and the close
+ * pipeline read the second as the first. A gate that reports success when it
+ * did not run is worse than no gate, because it is trusted.
+ *
+ * This module is that missing channel. A degradation travels **beside** the
+ * `Finding[]`, never inside it:
+ *
+ *   - it never becomes a `Finding`, so `countBySeverity` is untouched and no
+ *     execution failure can appear as a critical / high / medium (or even a
+ *     suggestion) — #4699's intent survives intact;
+ *   - it is rendered as its own section in the structured comment, and it
+ *     suppresses the false `✅ No findings` claim;
+ *   - it is carried on the `runCodeReview` envelope as `degraded` /
+ *     `degradations[]`, so the close pipeline sees it too.
+ *
+ * ## Report, not block — and why
+ *
+ * A degraded gate is reported loudly and does **not** halt the close. The close
+ * pipeline already runs the canonical `npm run lint` as a hard
+ * close-validation gate *before* the review phase; the review's scoped lint is
+ * a second, narrower read of the same surface. Failing a merge because a
+ * *secondary* read of an already-gated surface could not start would block
+ * delivery on an operational condition the hard gate has already covered. What
+ * was actually broken was the silence, so the fix is to make the silence
+ * impossible: every surface that reads the review outcome now states the
+ * degradation explicitly. Escalating to a block is a one-line change here if
+ * the operator posture ever needs it.
+ */
+
+/**
+ * @typedef {object} GateDegradation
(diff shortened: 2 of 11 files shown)


dsj1984:
## 🔬 Code Review — Story #4839

**Comparison**: `main` … `story-4839`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4852, opened 2026-07-30
fix(rollup): stop the follow-up roll-up asserting the cumulative friction ledger happened in the current plan-run (#4850)

### LINKED ISSUE #4850: fix(rollup): stop the follow-up roll-up asserting the cumulative friction ledger happened in the current plan-run
## Goal
Make a filed friction follow-up describe the corpus it was actually reduced over — a cross-run window of surviving signal streams — instead of claiming every occurrence happened in the plan-run that triggered the epilogue, and bound that window by age so a defect fixed weeks ago stops re-routing forever.

## Spec
The cross-run sweep is deliberate and MUST be preserved. gatherRunFrictionSignals in .agents/scripts/lib/orchestration/story-follow-ups.js reduces over every surviving per-Story stream because a systemic defect fires once per Story and scored 1 under the old per-run window, so it was discarded as a singleton forever. Narrowing the scan back to the run's own ids is refused.

The defect is that the rendering lies about that corpus. retro-proposals.js composes a title reading "recurred N times in <anchor>" and a body opening "surfaced N times during <anchor>", where <anchor> is the triggering run — while the evidence block in the same body already lists the foreign Stories the occurrences came from. The title is what a human triages on, so it must name the window: occurrence count, distinct Story count, date range. The triggering run stays in the body as a separate labelled fact, never as the scope of the claim.

Two constraints follow. The composer has no timestamp — normalizeGatheredSignal in .agents/scripts/lib/observability/runtime-friction.js drops the row's ts — so ts must be carried through that one shared normalizer, not re-read in a second place. And run-epilogue.js rewrites the anchor by regex over a "plan-run <number>" substring; once the composer stops emitting it, the run token must reach the composer as data.

The age floor fails toward under-counting, the direction that fails toward not filing. Rows with no parseable ts are excluded and counted, never dropped silently.

Idempotency is out of scope: the marker is category-fingerprint-only and anchor-free, and the title is not fingerprinted, so retitling cannot re-file the backlog.

## Changes
- `.agents/scripts/lib/orchestration/retro-proposals.js` — refactors-existing
- `.agents/scripts/lib/orchestration/story-follow-ups.js` — refactors-existing
- `.agents/scripts/lib/orchestration/run-epilogue.js` — refactors-existing
- `.agents/scripts/lib/observability/runtime-friction.js` — refactors-existing
- `.agents/scripts/lib/config-settings-schema-delivery.js` — refactors-existing
- `.agents/schemas/agentrc.schema.json` — refactors-existing
- `tests/lib/orchestration/retro-proposals.test.js` — refactors-existing
- `tests/lib/orchestration/story-follow-ups.test.js` — refactors-existing
- `tests/lib/observability/runtime-friction.test.js` — refactors-existing
- `tests/plan-run-epilogue.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: Given a gathered corpus whose rows carry storyIds outside the triggering run's id set, the composed proposal title states the occurrence count, the count of distinct contributing Stories, and the window's first-to-last date range, and contains no substring asserting the occurrences happened in the triggering run.
- [ ] AC-2: The issue body's opening line describes the corpus window rather than reading 'during <anchor>', and the triggering run appears as its own labelled evidence fact alongside the existing contributing-Stories list.
- [ ] AC-3: normalizeGatheredSignal carries each row's ts through to the composer's signal shape, resolving to null when the field is absent or unparseable, and remains the single normalizer both the story-scope and run-scope gathers call.
- [ ] AC-4: gatherRunFrictionSignals excludes rows older than delivery.feedbackLoop.frictionWindowDays from the recurrence window, defaulting to 30 days when the knob is unset; a row whose ts cannot be parsed is likewise excluded rather than aged in.
- [ ] AC-5: The follow-up-rollup step result reports the window bound that was applied and the number of rows excluded as stale or unparseable, so an operator can distinguish a bounded window from an unbounded one without reading the comment prose.
- [ ] AC-6: run-epilogue no longer rewrites proposal titles or bodies by regex over a 'plan-run' plus digits substring; the triggering run token reaches the composer as an input.
- [ ] AC-7: The runtime AJV in config-settings-schema-delivery.js accepts delivery.feedbackLoop.frictionWindowDays as an integer of at least 1 and rejects a non-integer or zero value, and the published mirror in .agents/schemas/agentrc.schema.json carries the same key.
- [ ] AC-8: A corpus confined to the triggering run's own Stories still renders a title and body that name that run, so the common single-run case does not regress into hedged wording.

## Verify
- node --test tests/lib/orchestration/retro-proposals.test.js (unit)
- node --test tests/lib/orchestration/story-follow-ups.test.js (unit)
- node --test tests/lib/observability/runtime-friction.test.js (unit)
- node --test tests/plan-run-epilogue.test.js (unit)
- npm run lint (validate)

## Non-Goals
- Scoping the signal scan back to the run's --stories set — that reverts the recurrence window Story #4824 widened and re-breaks the framework limb of the feedback loop.
- Changing the >= 2 actionable threshold, the dominant-source routing, or the recovery-netting.
- Changing the idempotency marker or its fingerprint — the title is already excluded from the hash.
- Purging or reaping signals.ndjson; the age floor governs window entry only and writes nothing.

dsj1984:
Closes #4850

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes follow-up filing thresholds and rendered operator-facing roll-up text in the delivery epilogue; behavior is well-covered by tests but mis-tuning `frictionWindowDays` could suppress legitimate recurring friction.
> 
> **Overview**
> Fixes follow-up roll-ups that treated the **cross-run friction ledger** as if every occurrence happened in the **current plan-run**, and stops **stale signals** from re-filing issues indefinitely.
> 
> **Age-bounded gather:** `gatherRunFrictionSignals` now drops rows older than `delivery.feedbackLoop.frictionWindowDays` (default **30**) or with no parseable `ts`, counts those exclusions, and returns `{ signals, window }` instead of a bare array. Timestamps flow through the shared `normalizeGatheredSignal` path so story- and run-scope gathers stay aligned.
> 
> **Honest proposal copy:** The epilogue passes **`runToken`** and **`anchorStoryIds`** into `composeRoutedProposals` and removes the post-hoc **`plan-run \d+` regex rewrite**. Cross-window corpora are titled by occurrence count, Story span, and date range, with the triggering run labeled separately; run-confined corpora keep the familiar “in plan-run …” wording.
> 
> **Operator visibility:** Follow-up rollup step results now include **`frictionWindow`** (days, cutoff, excluded stale/unparseable counts). Config is documented in agentrc schema, runtime AJV, and generated docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b45e7ba744aa4aa81cda21b247222666e1a73ecd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4850

**Comparison**: `main` … `story-4850`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 1
- 🟢 Suggestion: 0

### 🟡 Medium Risk (1)

#### 🟡 Size/Volume Warning — `.agents/scripts/lib/config-settings-schema-delivery.js` _[maintainability]_

Module `.agents/scripts/lib/config-settings-schema-delivery.js` reports a size/volume warning (module 44.8).

PR #4868, opened 2026-07-31
fix: stop the CRAP row builder silently mixing transpiled and original line coordinates, and refuse to compare on an unsound basis (#4866)

### LINKED ISSUE #4866: Stop the CRAP row builder silently mixing transpiled and original line coordinates, and refuse to compare on an unsound basis
## Goal
Make a CRAP row's line coordinate system explicit so the gate can no longer mis-key a method against the baseline and report a regression no change can satisfy.

## Spec
**One coordinate system per row, recorded.** `methodRowsFromReport` currently substitutes the raw transpiled line whenever `mapLine` returns null, so a single scan can emit rows in two coordinate systems with nothing on disk distinguishing them. A row must instead carry its provenance — remapped-to-original vs un-remapped-transpiled — and an un-remapped row must not be silently treated as an original-source coordinate, because that same coordinate is what the coverage join reads.

**Refuse, don't invent.** Rows whose coordinate provenance differs from the baseline's are not comparable; the compare must say so rather than resolving them through the nearest-line drift heuristic, which is what manufactures the phantom.

**Unsound-basis backstop.** Above a drifted-row ratio threshold the comparison basis is self-evidently unsound. The preview gate then stops emitting per-method verdicts, prints one named diagnostic telling the operator to re-seed, and **exits 0** — failing open. The authoritative `check-baselines` gate still gates the merge, so failing the pre-commit hook open costs no real coverage while blocking it costs a provably-defect-free commit.

**Close the compat door.** The preview gate evaluates the loaded-envelope compatibility check before it compares anything, and the TS-transpiler coordinate axis becomes reachable in production rather than dead — a transpiler change moves every TS row's coordinate, which is half the identity key.

**Invariant.** Pure-JavaScript scans are unaffected: their coordinates already are original-source coordinates.

## Changes
- `.agents/scripts/lib/crap-engine.js` — refactors-existing
- `.agents/scripts/lib/baselines/preview-gates.js` — refactors-existing
- `tests/lib/crap-engine.test.js` — refactors-existing
- `tests/lib/crap-utils.sourcemap-join.test.js` — refactors-existing
- `tests/baselines/crap-compat-axes.test.js` — refactors-existing
- `tests/quality-preview.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: A CRAP row records which line coordinate system its startLine is in, so a row whose sourcemap lookup did not resolve is distinguishable on disk from one that did.
- [ ] AC-2: A method whose generated line has no sourcemap entry is no longer emitted with a transpiled startLine presented as an original-source coordinate.
- [ ] AC-3: A fixture scan of a TS/TSX source in which some methods resolve through the sourcemap and others do not produces rows in exactly one coordinate system, or reports the unresolved ones as such — never a silent mix.
- [ ] AC-4: The compare refuses to resolve rows whose coordinate provenance differs from the baseline's through the nearest-line drift heuristic.
- [ ] AC-5: When the drifted-row ratio exceeds the unsound-basis threshold, the preview gate emits one named diagnostic naming the coordinate mismatch and the re-seed remedy, suppresses its per-method verdicts, and exits 0.
- [ ] AC-6: runCrapPreview evaluates the loaded-envelope baseline-compatibility check before comparing, and an incompatible baseline yields that diagnostic rather than per-method regressions.
- [ ] AC-7: The TS-transpiler coordinate compat axis is reachable from production code rather than tests only, and a transpiler-version change is treated as a coordinate-invalidating condition rather than an advisory warning.
- [ ] AC-8: Scanning a pure-JavaScript tree produces byte-identical rows and the same crap values as before the change, and the committed CRAP baseline needs no refresh to stay green.

## Verify
- node --test tests/lib/crap-engine.test.js (unit)
- node --test tests/lib/crap-utils.sourcemap-join.test.js (unit)
- node --test tests/baselines/crap-compat-axes.test.js (unit)
- node --test tests/quality-preview.test.js (unit)
- npm run crap:check (contract)
- npm test (contract)
- npm run lint (validate)

## Non-Goals
- Re-keying escomplex's `<anon method-N>` ordinal, or replacing it with a content or structural key — the ordinal is the upstream analyzer's and is stable for identical source.
- Merging quality-preview.js and update-crap-baseline.js onto a shared keying helper; they already share one.
- Changing the CRAP formula, the per-method ceiling, or any committed baseline value for a pure-JavaScript scan.
- Making the authoritative check-baselines CRAP gate fail open — the backstop is scoped to the preview gate.

dsj1984:
Closes #4866

_Auto-opened by `/deliver`._


dsj1984:
## 🔬 Code Review — Story #4866

**Comparison**: `main` … `story-4866`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 1
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### 🟠 High Risk (1)

#### 🟠 Lint check failed (1 error(s)) _[lint]_

Scoped lint reported 1 error(s) and 0 warning(s) on the changed surface. Fix errors before merging.

PR #4881, opened 2026-08-01
fix: make the CRAP preview gate report what it measures: fix the unsound-basis numerator and stop scoring absent coverage as zero (#4871)

### LINKED ISSUE #4871: Make the CRAP preview gate report what it measures: fix the unsound-basis numerator and stop scoring absent coverage as zero
## Goal
Restore the CRAP preview gate's per-method verdicts by counting the condition the backstop was built to detect and by declining to score coverage it was never given.

## Spec
**The numerator is wrong, not miscalibrated.** `assessComparisonBasis` suppresses every per-method verdict when the drifted-plus-incomparable ratio exceeds its threshold, but `drifted` is counted after the provenance filter has already removed provenance-mismatched rows. The quantity it reads can therefore never evidence the coordinate mixing the backstop exists to catch, and on a normal diff the ratio trips anyway — so the preview gate returns exit code 0 before regressions are examined and reports nothing.

**Absent coverage is not zero coverage.** A story worktree is created without a coverage directory, and an unresolved method is then scored at zero percent covered, so CRAP goes maximal and the first commit of most Stories fails on files it never touched. Those unscorable rows also inflate the denominator of the same ratio, so the two defects compound. A method with no coverage artifact must be reported unscorable rather than computed from an assumed zero.

**Preserve the deliberate fail-open.** Suppressing verdicts when the basis is genuinely unsound is the shipped design and stays; the authoritative gate still gates the merge. What changes is that suppression fires only on real provenance-mismatch evidence, captured before the filter removes it.

**Invariant.** A pure-JavaScript scan, whose coordinates are already original-source, must yield per-method verdicts on every diff — that is the regression tell.

## Changes
- `.agents/scripts/lib/baselines/preview-gates.js` — refactors-existing
- `.agents/scripts/lib/crap-utils.js` — refactors-existing
- `tests/quality-preview.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: A pure-JavaScript diff that introduces a real per-method CRAP regression produces a per-method verdict from the preview gate rather than a suppressed run.
- [ ] AC-2: The unsound-basis determination is derived from provenance-mismatch evidence captured before the provenance filter removes it, so the quantity compared against the threshold can actually observe coordinate mixing.
- [ ] AC-3: A fixture scan with genuinely mixed coordinate provenance still suppresses per-method verdicts, emits one named diagnostic identifying the mismatch, and exits 0.
- [ ] AC-4: A fixture scan with uniform coordinate provenance never suppresses per-method verdicts regardless of how many rows drift.
- [ ] AC-5: With no coverage artifact present, a method is reported as unscorable rather than scored from an assumed zero percent coverage.
- [ ] AC-6: An unscorable row does not contribute to the ratio that decides whether the comparison basis is unsound.
- [ ] AC-7: A commit made in a freshly initialized story worktree with no coverage directory does not fail the pre-commit CRAP gate on files it did not modify.
- [ ] AC-8: An operator can distinguish a suppressed-because-unsound run from a clean run by the gate's own output alone.

## Verify
- node --test tests/quality-preview.test.js (unit)
- node .agents/scripts/quality-preview.js --staged (contract)
- npm test (contract)
- npm run lint (validate)

## Non-Goals
- Removing the unsound-basis backstop or making the preview gate fail closed — the fail-open is the shipped design from the coordinate-provenance work.
- Re-tuning the ratio threshold or the minimum sample size as the primary fix; the count feeding them is the defect.
- Changing the CRAP formula, the per-method ceiling, or any committed baseline value.
- Altering the authoritative check-baselines CRAP gate, or making a coverage run part of worktree initialization.

dsj1984:
Closes #4871

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes quality-gate comparison logic and scoring when coverage is absent; behavior is heavily tested but affects pre-commit preview and baseline compare paths developers rely on.
> 
> **Overview**
> Fixes the CRAP **preview gate** so its unsound-comparison backstop measures **coordinate provenance mixing**, not ordinary line drift, and stops treating a **missing coverage run** as 0% tested code.
> 
> **Unsound-basis ratio (#4871).** `compareCrap` now records **`provenanceMismatched`** before the provenance filter (any baseline candidate in a different line coordinate system). `assessComparisonBasis` uses that count (or `incomparable`) over **`comparable`** rows instead of `drifted + incomparable` over `total`. Pure-JS diffs where every method drifts after an insertion no longer falsely suppress per-method verdicts; genuine coordinate mixing still triggers `crap-unsound-comparison-basis` and fails open on preview.
> 
> **Unscorable methods.** Rows with `crap`/`coverage` null (or `unscorable`) are bucketed separately, excluded from compare arms and from the backstop denominator. **`finalizeMethodRows`** and the scan/worker path honor **`coverageAvailable`**: when no `coverage-final.json` was loaded, unresolved methods are skipped and counted—not filled with 0% CRAP (`c² + c`), which was failing first commits in fresh worktrees.
> 
> Report JSON schema and envelopes gain **`provenanceMismatched`** and **`unscorable`**; CLI summary printing reflects both counters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a974e3199ac1d123450152037db656d16109693. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4871

**Comparison**: `main` … `story-4871`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4883, opened 2026-08-01
fix: stamp the version cache with the real clock and stop the env loader fusing inline comments into values (#4878)

### LINKED ISSUE #4878: Stamp the version cache with the real clock and stop the env loader fusing inline comments into values
## Goal
Fix two mandrel CLI defects that each corrupt state the tool itself later reads.

## Spec
**The bypass poisons the cache it bypasses.** The update command's cache bypass is implemented by shifting the clock forward 48 hours so any existing entry reads as stale, but that same shifted clock is then threaded into the cache write, stamping the freshness timestamp roughly two days into the future. Every successful update re-poisons it, and the 24-hour freshness window is defeated until real time catches up. The staleness decision may use whatever clock it needs, but the persisted timestamp must record when the probe actually happened.

**A value ends where the value ends.** The environment loader matches a key and then captures the remainder of the line, stripping only surrounding quotes, so a trailing inline comment is fused into the value. A token line carrying a trailing note becomes an invalid credential that fails authentication with no indication the comment is the cause. An unquoted value must end at an unescaped comment marker; a quoted value must keep its contents verbatim.

## Changes
- `lib/cli/update.js` — refactors-existing
- `lib/cli/version-check.js` — refactors-existing
- `.agents/scripts/lib/env-loader.js` — refactors-existing
- `tests/lib/env-loader.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: After an explicit update, the persisted version-cache timestamp is not in the future relative to the real clock.
- [ ] AC-2: An explicit update still performs exactly one registry probe rather than reusing a fresh cache entry.
- [ ] AC-3: A passive staleness check performed within 24 hours of an explicit update reuses the cache instead of probing again.
- [ ] AC-4: An unquoted environment value followed by an inline comment loads without the comment text.
- [ ] AC-5: A quoted environment value containing a comment marker keeps that marker in its value.
- [ ] AC-6: A token supplied with a trailing inline comment authenticates successfully.

## Verify
- node --test tests/lib/env-loader.test.js (unit)
- npm test (contract)
- npm run lint (validate)

## Non-Goals
- Changing the 24-hour freshness window or adding a cache-control flag to the update command.
- Replacing the hand-rolled environment loader with a third-party parser.
- Changing which files the loader reads or their precedence.

dsj1984:
Closes #4878

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes affect how `.env` values (often secrets) are parsed and how the CLI version cache behaves after `mandrel update`; behavior is covered by new unit tests but both paths are operationally sensitive.
> 
> **Overview**
> Fixes **#4878** by stopping `mandrel update` from writing a **future** `checkedAt` into the version freshness cache. Cache bypass no longer advances `now` by ~48h; `isStale` gains **`forceRefresh`** so explicit updates still probe the registry while **`checkedAt` uses the real clock**, so passive checks within 24h reuse the cache again.
> 
> Separately, **`.env` loading** now parses values through **`parseEnvValue`**: trailing **`#` comments** are stripped on unquoted values, **`#` inside quotes** is kept, and **`\#`** can emit a literal hash—avoiding tokens or secrets being corrupted when someone adds an inline note after `=`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed564c86fc62d80ddfaf1f3f88be429ba8f0bc42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/env-loader.js
+++ b/.agents/scripts/lib/env-loader.js
@@ -3,6 +3,51 @@ import path from 'node:path';
 import { Logger } from './Logger.js';
 
 /**
+ * Parse the right-hand side of a `KEY=` line into its value.
+ *
+ * Two rules, and the quoting decides which one applies:
+ *
+ *  - **Quoted** (`"…"` / `'…'`): the contents between the opening quote and
+ *    its matching closing quote are kept **verbatim** — a `#` inside quotes is
+ *    part of the value, not a comment. Anything after the closing quote is
+ *    trailing commentary and is discarded.
+ *  - **Unquoted**: the value ends at the first **unescaped** `#`. A `\#`
+ *    escape emits a literal `#` and keeps the scan going, so a value that
+ *    genuinely contains a hash can still be written without quotes.
+ *
+ * An opening quote with no closing partner is not a quoted value — it falls
+ * through to the unquoted rule so a malformed line still yields something
+ * rather than swallowing the rest of the file's intent.
+ *
+ * @param {string} raw — the text after the `=`, unparsed.
+ * @returns {string} The value with any inline comment removed.
+ */
+function parseEnvValue(raw) {
+  const trimmed = raw.trim();
+  if (trimmed.length === 0) return '';
+
+  const quote = trimmed.charAt(0);
+  if (quote === '"' || quote === "'") {
+    const closing = trimmed.indexOf(quote, 1);
+    // Verbatim contents: no comment stripping, no unescaping.
+    if (closing !== -1) return trimmed.slice(1, closing);
+  }
+
+  let value = '';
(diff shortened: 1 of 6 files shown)


dsj1984:
## 🔬 Code Review — Story #4878

**Comparison**: `main` … `story-4878`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4894, opened 2026-08-01
fix: pin the base branch at init and fail closed at close when config has changed mid-run (#4891)

### LINKED ISSUE #4891: Pin the base branch at init and fail closed at close when config has changed mid-run
## Goal
A Story must be base-synced against the same base branch it was seeded from, and a config change during the implementation window must surface as an explicit diagnostic rather than as an unrelated content conflict.

## Spec
single-story-init.js resolves config and seeds story-<id> from project.baseBranch; the close runner in lib/orchestration/single-story-close/runner.js re-resolves the same value and runs close-validation, format-autofix, the gate baseline and base-sync against **its** answer. Nothing pins the first value or checks the two agree. The window is unbounded — it is however long implementation takes. The existing lease-based cross-run hardening cannot help: the contended resource is not another run, it is a mutable file read twice.

The value is already recorded. init upserts a story-init structured comment whose JSON payload carries baseBranch, and findStructuredComment in lib/orchestration/ticketing/reads.js is the existing read seam. Prefer that receipt over the temp log the report proposed: the post-land tail purges the temp tree, while the comment survives, which dissolves the reported edge case. The comment upsert is best-effort, so absence must still be handled.

Invariants:

- Close consults the run's pinned base, not a fresh resolution. The base a Story was seeded from is a property of that run.
- Disagreement between the pinned value and current config **fails closed**, naming both values. Never silently prefer either. This is the diagnostic that was missing.
- Base-sync conflict remediation must not advise merging the base until the base is confirmed to match the receipt.
- A missing receipt falls back to resolving config **with an explicit warning**. Silent fallback reintroduces the bug.

Pin the value through a seam that generalizes to other run-scoped config, even though only the base branch is enforced here.

## Changes
- `.agents/scripts/single-story-init.js` — refactors-existing
- `.agents/scripts/lib/orchestration/single-story-close/runner.js` — refactors-existing
- `tests/single-story-close-sync.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: Close derives the base branch from the run's init receipt rather than re-resolving project.baseBranch, so a Story seeded from one base is base-synced against that same base even when config changed after init.
- [ ] AC-2: When the receipt's base branch and the currently-resolved base branch disagree, close refuses with an explicit config-changed-mid-run error that names both values, and performs no base-sync, format-autofix or gate run.
- [ ] AC-3: A base-sync conflict no longer emits remediation advising the operator to merge the base until the base has been confirmed to match the receipt.
- [ ] AC-4: When the init receipt is absent (a close re-run after the temp tree was purged, or a recovery path), close falls back to resolving config and emits an explicit warning naming the fallback; the fallback is never silent.
- [ ] AC-5: The pinning seam records the run-scoped value in a form that other config keys can be added to without a second mechanism.

## Verify
- node --test tests/single-story-close-sync.test.js (unit)
- node .agents/scripts/single-story-init.js --help (validate)

## Non-Goals
- Snapshotting and enforcing every run-scoped config value (ceremony profile, quality floors, concurrency cap) — the seam should generalize, but only the base branch is enforced in this Story because only its corruption is destructive.
- Changing the lease or cross-run de-confliction machinery, which is a different class and was not engaged in the observed failure.
- Preventing concurrent sessions from editing .agentrc.

dsj1984:
Closes #4891

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes pre-push close behavior and git remediation guidance for base-sync failures—wrong handling could block closes or mislead recovery, but the fail-closed design reduces risk of merging the wrong base.
> 
> **Overview**
> **Pins `project.baseBranch` at story init and enforces it at close** so a Story is never base-synced or remediated against a base it was not seeded from if `.agentrc` changes mid-run.
> 
> Init records `runScopedConfig` (via new `pinRunScopedConfig`) in the ticket’s `story-init` structured comment. Close calls `resolveRunScopedConfig` **before** gates, format-autofix, and base-sync: if the receipt disagrees with current config it throws with both values; if the receipt is missing or unreadable it falls back to current config with an explicit warning and `confirmed: false`.
> 
> Base-sync failure friction comments now take `baseConfirmed` (default false): **`git merge origin/<base>` recovery steps are only shown when the base was confirmed** against the init receipt; otherwise operators are told to establish the real base from the `story-init` comment first.
> 
> `resolveRunScopedConfig` is re-exported from `single-story-close.js` for tests; coverage is expanded in `single-story-close-sync.test.js`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e403ebf9063f9b080a7b1e2f5ad417e5c4478413. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/orchestration/run-scoped-config.js
+++ b/.agents/scripts/lib/orchestration/run-scoped-config.js
@@ -1,0 +1,276 @@
+/**
+ * run-scoped-config.js — pin the config values a delivery run was seeded
+ * from, and fail closed when the file they came from changed underneath it.
+ *
+ * The problem this exists for: `single-story-init.js` resolves config and
+ * seeds `story-<id>` from `project.baseBranch`; `single-story-close` used to
+ * re-resolve the SAME key and run close-validation, format-autofix, the gate
+ * baseline and base-sync against its own second answer. Nothing pinned the
+ * first value and nothing checked the two agreed — and the window between
+ * them is however long implementation takes. A concurrent session editing
+ * `.agentrc.json` / `.agentrc.local.json` mid-run therefore base-synced a
+ * Story against a base it was never seeded from, which surfaced only as an
+ * ordinary content conflict with nothing pointing at config.
+ *
+ * The base a Story was seeded from is a property of **that run**, not of
+ * whatever the config file says later. So init pins it and close reads the
+ * pin back.
+ *
+ * ## Where the pin lives
+ *
+ * The `story-init` structured comment, not a temp log: the post-land tail
+ * purges the temp tree, while the ticket comment survives every close re-run
+ * and every recovery path. `single-story-init.js` already upserts that
+ * comment; `pinRunScopedConfig` supplies the block it records, and
+ * `resolveRunScopedConfig` reads it back through the existing
+ * `findStructuredComment` seam.
+ *
+ * ## Adding another run-scoped key
+ *
+ * Add one row to `RUN_SCOPED_CONFIG_KEYS`. Both halves — the pin init writes
+ * and the comparison close makes — enumerate that registry, so a new key
+ * needs no second mechanism, no reader change and no writer change. Only
+ * `baseBranch` is enforced today because only its corruption is destructive
+ * (a wrong base merged into the Story branch permanently contaminates the
+ * branch and its PR diff); ceremony profile, quality floors and the
+ * concurrency cap are deliberately NOT pinned here.
+ */
+
+import { parseFencedJsonComment } from './structured-comment-parser.js';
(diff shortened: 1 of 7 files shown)


dsj1984:
## 🔬 Code Review — Story #4891

**Comparison**: `main` … `story-4891`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4908, opened 2026-08-02
docs: give deliver-digest section 3 the actual call shapes for deriveChangeLevel and resolveCeremonyForRisk, and guard them against drift (#4904)

### LINKED ISSUE #4904: Give deliver-digest section 3 the actual call shapes for deriveChangeLevel and resolveCeremonyForRisk, and guard them against drift
## Goal
Make the digest's ceremony-routing incantation runnable as written, and add an executable guard so a future signature change cannot leave it silently wrong.

## Spec
Section 3 of .agents/workflows/helpers/deliver-digest.md is the single bundled read a deliverer uses to derive ceremony. It names `deriveChangeLevel` and `resolveCeremonyForRisk` and describes their semantics correctly, but never states their argument or return shapes. The real contracts are `deriveChangeLevel({ changedFiles, injectedRules?, selectSensitivePathClassesFn? }) -> { level, classes }` and `resolveCeremonyForRisk({ derivedLevel, clusterIndex?, freshCriticSampleRate?, ceremonyProfile? }) -> { mode, reason, sampled, profile, verdictOwner }`.

Two things make the omission costly rather than cosmetic. First, `deriveChangeLevel` returns an OBJECT, while the section-3 phrasing in .agents/workflows/helpers/deliver-digest.md ("Derive the level with `deriveChangeLevel` over that one list") reads as if it returns the level — so the natural composition passes `{ derivedLevel: <object> }` where a level string belongs. Second, both functions are documented total and never throw, and an unrecognized level resolves to the `null` fail-safe, which routes to the MORE thorough posture. The misuse therefore costs a fresh critic sub-agent on every low-risk Story and emits no error at all — it fails expensively and invisibly. A consumer deliverer hit exactly this and reported it as friction.

The fix is the shapes stated concretely in section 3 of .agents/workflows/helpers/deliver-digest.md, plus a guard that cannot rot: assert by EXECUTION, not by string-matching the prose. Calling the functions with the documented shapes must yield a real level and a real mode, and calling them with the level-not-object shorthand must yield the `null` fail-safe — which is what proves the documented shape is load-bearing. A test that only greps the markdown for key names would pass against a renamed parameter.

`tests/workflows/deliver-digest.test.js` already asserts a digest byte budget. Any addition must stay inside it, or raise it deliberately with the new value and the reason stated — silently growing the budget defeats the read-cost discipline the digest exists to enforce.

## Changes
- `.agents/workflows/helpers/deliver-digest.md` — refactors-existing
- `tests/workflows/deliver-digest.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: `deliver-digest.md` § 3 states the argument shape of both functions with their real key names (`changedFiles`, `derivedLevel`) and makes explicit that `deriveChangeLevel` returns `{ level, classes }` rather than a bare level, so the composition between the two calls is unambiguous from the digest alone.
- [ ] AC-2: A test in `tests/workflows/deliver-digest.test.js` calls `deriveChangeLevel` and `resolveCeremonyForRisk` with the shapes the digest documents and asserts a non-null `level` and a concrete `mode` come back — the digest's incantation is verified by execution, not by matching its prose.
- [ ] AC-3: The same test asserts the negative control: passing the result of `deriveChangeLevel` directly as `derivedLevel` (the object-for-string mistake the old wording invites) resolves to the `null`-level fail-safe posture rather than a `low` verdict. This is what proves the documented shape is load-bearing and would catch a future parameter rename.
- [ ] AC-4: The digest stays inside the byte budget its existing test enforces; if the budget is raised, the commit body states the new value and why the addition earns it.
- [ ] AC-5: `npm test` and `npm run lint` both pass, including the pre-existing `deliver-digest.md` assertions for surface coverage, spine reachability, and the terminal-table citation.

## Verify
- npm test (unit)
- npm run lint (validate)

## Non-Goals
- Changing `deriveChangeLevel` or `resolveCeremonyForRisk` themselves — their signatures, their totality, and the `null`-routes-to-thorough fail-safe are all correct and stay as they are. This Story fixes the documentation of them and guards it.
- Restating the same shapes in `deliver-story.md`, `deliver-reference.md`, or `plan-reference.md`. § 3 is the single home; duplicating it creates the drift this Story is closing.
- Revisiting the ceremony routing policy itself (`minimal` always inline, `strict` always fresh, `standard` by derived level, the `freshCriticSampleRate` floor).

dsj1984:
Closes #4904

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Low Risk**
> Documentation and workflow test changes only; no runtime orchestration or delivery logic is modified.
> 
> **Overview**
> **`deliver-digest.md` §3** now documents the real call contract for change-set ceremony routing: `deriveChangeLevel({ changedFiles })` returns **`{ level, classes }`**, and `resolveCeremonyForRisk` must receive **`derivedLevel` as the level string**—not the whole object. The section adds a runnable `node -e` snippet that composes `computeChangeSet`, both routing helpers, and JSON output, and explains that both APIs are **total (never throw)**, so mis-shaped arguments silently hit the **`null` fail-safe** (extra fresh critics) instead of errors.
> 
> **`deliver-digest.test.js`** extends digest coverage checks for those argument/return tokens and adds **#4904 execution tests** that call the documented composition against real modules: correct shapes yield a concrete mode and `verdictOwner`, and passing the `{ level, classes }` object as `derivedLevel` must route **`fresh`** with an **underivable** reason while the string form stays **`inline`** on pinned `low` risk.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35387bc7576024694d53568a2d04864832449f3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/workflows/helpers/deliver-digest.md
+++ b/.agents/workflows/helpers/deliver-digest.md
@@ -54,21 +54,32 @@ the only sanctioned landing. A silent local build is not a delivery.
 ## 3. Change set — computed once, handed to everyone
 
 One enumeration per Story. A critic that re-runs its own `git diff`
-can score a different set than the one that routed it:
+can score a different set than the one that routed it. Both routing calls take
+a single options object and are **total — they never throw**, so a wrong-shaped
+argument is absorbed into the `null` fail-safe and reports nothing:
 
 ```bash
 node --input-type=module -e '
-  import { computeChangeSet } from "<main-repo>/.agents/scripts/lib/orchestration/change-set.js";
+  const lib = "<main-repo>/.agents/scripts/lib/orchestration";
+  const { computeChangeSet } = await import(`${lib}/change-set.js`);
+  const { deriveChangeLevel } = await import(`${lib}/review-depth.js`);
+  const { resolveCeremonyForRisk } = await import(`${lib}/ceremony-routing.js`);
   const { files } = computeChangeSet({ baseRef: "main", headRef: "story-<storyId>" });
-  console.log(JSON.stringify(files));
+  // deriveChangeLevel({ changedFiles, injectedRules?, selectSensitivePathClassesFn? })
+  //   -> { level, classes } — an OBJECT, never a bare level.
+  const { level, classes } = deriveChangeLevel({ changedFiles: files });
+  // resolveCeremonyForRisk({ derivedLevel, clusterIndex?, freshCriticSampleRate?,
+  //   ceremonyProfile? }) -> { mode, reason, sampled, profile, verdictOwner }.
+  // derivedLevel is that level STRING. Handing it the object above matches no
+  // tier, so it routes to the null fail-safe: a fresh critic, silently.
+  const ceremony = resolveCeremonyForRisk({ derivedLevel: level, clusterIndex: 0 });
+  console.log(JSON.stringify({ files, level, classes, ...ceremony }));
 '
 ```
 
-Derive the level with `deriveChangeLevel`
-([`review-depth.js`](../../scripts/lib/orchestration/review-depth.js)) over
-that one list: a sensitive path registered in `audit-rules.json` → `high`, none
-→ `low`, an unenumerable diff (`files === null`) → `null`. Resolve
-fresh-vs-inline critics with `resolveCeremonyForRisk`
+Level rules ([`review-depth.js`](../../scripts/lib/orchestration/review-depth.js)):
+a sensitive path registered in `audit-rules.json` → `high`, none → `low`, an
+unenumerable diff (`files === null`) → `null`. Ceremony rules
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4904

**Comparison**: `main` … `story-4904`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4917, opened 2026-08-02
fix: stop check-baselines silently disabling its compare arm when the base baseline cannot be read (#4914)

### LINKED ISSUE #4914: Stop check-baselines silently disabling its compare arm when the base baseline cannot be read
## Goal
A base baseline larger than 1 MB kills `git show` with ENOBUFS and check-baselines reports zero regressions and zero additions at exit 0 — make the read succeed, and make any remaining read failure red the gate instead of quietly emptying it.

## Spec
Two defects compose into a fail-open gate.

**Trigger.** `readBaseFromGit` calls `_spawnSync('git', ['show', ref:file])` with no `maxBuffer`, inheriting Node's 1 MB default. A larger base baseline kills the child (`status: null`, `signal: SIGTERM`, `error.code: ENOBUFS`) and the function throws. Measured against swarm-os `main` (crap baseline 1,178,910 bytes): default → `status=null signal=SIGTERM err=ENOBUFS bytes=1064960`; at 64 MB → `status=0 bytes=1178910`. `readRangeSubjectsTouchingFile` has the same missing-`maxBuffer` shape and returns `[]` on any failure, which silently un-acknowledges a `baseline-refresh:` commit tag. Both sites take the 64 MB bound already used at `run-test-profile.js:87`, `audit-baselines/trend.js:61` and `weights.js:65` — git-base is the outlier, not the precedent-setter.

**Amplifier.** `readBaseBaselinePayload` wraps the call in a bare `catch { return null }`, so `evaluateCompare` returns `emptyCompareResult` and the whole head-vs-base arm — regressions AND additions — stops running while the floors arm keeps the gate at exit 0. This is the load-bearing constraint: `readBaseFromGit` already distinguishes the two cases itself, returning `null` **only** for git exit 128 (path absent at ref) and throwing on everything else. The throw is therefore unambiguously a hard failure and MUST NOT be conflated with an absent path. It fails closed as `EXIT_CONFIG` (3) — "the gate could not even start", the same code `assertFloorAxesExist` already uses for a misconfigured floor axis.

**Observability.** `baseRead` exists but never leaves the module (`compare.js:41`/`:68`, consumed at `:101`), which is why a dead compare arm looks byte-identical to a clean run. It must reach the per-gate report object so the condition is diagnosable from the JSON alone.

A JSON-parse failure on a successfully-read payload keeps its current degrade-to-null behavior; only the read failure changes.

## Changes
- `.agents/scripts/lib/baselines/git-base.js` — refactors-existing
- `.agents/scripts/lib/orchestration/check-baselines/phases/compare.js` — refactors-existing
- `.agents/scripts/lib/orchestration/check-baselines/phases/report.js` — refactors-existing
- `tests/baselines/git-base.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: AC-1: Both `_spawnSync` call sites in `.agents/scripts/lib/baselines/git-base.js` pass an explicit `maxBuffer` of `64 * 1024 * 1024`, matching the bound already used at `run-test-profile.js:87`.
- [ ] AC-2: AC-2: A unit test in `tests/baselines/git-base.test.js` injects a mock spawn via `__setSpawnRunner({spawn})` that returns the ENOBUFS shape (`status: null`, `signal: 'SIGTERM'`, `error.code: 'ENOBUFS'`) and asserts `readBaseFromGit` throws rather than returning null — locking the distinction the compare phase now depends on.
- [ ] AC-3: AC-3: A unit test asserts `readBaseFromGit` returns `null` (does not throw) when the mock spawn reports git exit 128, so the absent-path contract is proven still intact alongside AC-2.
- [ ] AC-4: AC-4: A unit test asserts both spawn sites received `maxBuffer: 67108864` in their options object, read off the mock's recorded argv/options rather than by inspecting source text.
- [ ] AC-5: AC-5: `readBaseBaselinePayload` no longer swallows a throw from `readBaseFromGit`: a hard read failure propagates as an error carrying `exitCode` `EXIT_CONFIG` (3), and a test asserts a check-baselines run whose base read throws exits 3 rather than 0.
- [ ] AC-6: AC-6: A test asserts the pre-fix scenario is genuinely closed end-to-end — with a mock spawn that ENOBUFS on the base read, the run does NOT report `regressions: []` / `additions: []` at exit 0.
- [ ] AC-7: AC-7: The per-gate report object emitted by check-baselines carries the compare arm's read status (`baseRead`, currently internal to compare.js), and a test asserts it is present and `true` on a normal run where the base baseline was read.
- [ ] AC-8: AC-8: `readRangeSubjectsTouchingFile` passes the same `maxBuffer`, and a test asserts it still returns `[]` (never throws) on a non-zero git status, so the refresh-tag path keeps its tolerant contract.
- [ ] AC-9: AC-9: An integration leg in `tests/baselines/git-base.integration.test.js` reads a real file larger than 1 MB at a real ref through actual git and asserts the full byte length is returned — the leg that would have caught this defect originally.
- [ ] AC-10: AC-10: `npm test` and `npm run lint` pass on the final tree.

## Verify
- node --test tests/baselines/git-base.test.js (unit)
- node --test tests/baselines/git-base.integration.test.js (contract)
- npm test (unit)
- npm run lint (validate)

## Non-Goals
- Changing what the compare arm does once the base payload IS readable — the classifiers, tolerance and refresh-acknowledgment logic are untouched.
- Shrinking any baseline file. The 1 MB ceiling is the bug; a baseline legitimately grows past it and must still be read.
- Reworking the `catch` around `JSON.parse` of a successfully-read payload. A malformed base baseline is a different failure with its own compat axis.
- Adding a newMethodCeiling arm to check-baselines. The per-method ceiling lives in close-validation's projectCrapBreaches by design; conflating the two is a separate discussion.
- Auditing every other spawnSync site in the tree for missing maxBuffer. This Story fixes the two in git-base.js that produced the observed failure.

dsj1984:
Closes #4914

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes check-baselines exit behavior when base reads fail (now exit 3 instead of a misleading pass), which affects CI quality gates but is intentional and heavily tested.
> 
> **Overview**
> Fixes a **fail-open** bug where oversized or broken base baseline reads were treated like “no base at ref,” so the compare arm reported zero regressions while floors still passed.
> 
> **`git-base.js`** sets an explicit **64 MB** `maxBuffer` on `git show` and `git log` (matching other baseline tooling), preventing ENOBUFS kills on baselines larger than spawnSync’s 1 MB default. The contract is documented: **`null` only for git exit 128** (path absent); signal kills and other failures **throw**.
> 
> **`compare.js`** no longer catches those throws and returns null; it surfaces **`EXIT_CONFIG` (3)** via `buildBaseReadError`. **Absent** bases (exit 128) still skip compare with exit 0.
> 
> Reports now expose **`baseRead`** on each gate (JSON) and **`[baseRead=false — compare skipped]`** in the text summary when compare did not run. New unit, integration, and end-to-end tests lock ENOBUFS behavior, buffer size, and CLI exit 3 vs tolerant absent-base exit 0.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a03ef2ac2d3f5ee32e48a42aab08aefdc3015a3b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/baselines/git-base.js
+++ b/.agents/scripts/lib/baselines/git-base.js
@@ -29,10 +29,15 @@
 //     loader treats this as a normal "no baseline at base ref" case
 //     and returns `null`. Callers branch on `result === null` rather
 //     than catching an error.
-//   - Any other non-zero exit (bad ref, corrupted repo, git binary
-//     missing): rethrown so the dispatcher can surface a config-error
-//     exit code rather than silently treating the failure as "no
-//     baseline". A bad ref is a config bug, not a missing file.
+//   - Any other outcome (bad ref, corrupted repo, git binary missing, or
+//     the child killed by signal so `status` is `null`): thrown so the
+//     dispatcher can surface a config-error exit code rather than
+//     silently treating the failure as "no baseline". A bad ref is a
+//     config bug, not a missing file.
+//
+// That two-way split is load-bearing (Story #4914): `null` means and only
+// means "path absent at ref". Callers may therefore treat a throw as a
+// hard failure without having to re-diagnose it.
 
 import { spawnSync } from 'node:child_process';
 
@@ -46,6 +51,21 @@ import { spawnSync } from 'node:child_process';
 // ---------------------------------------------------------------------------
 
 const DEFAULT_MAX_ENTRIES = 64;
+
+/**
+ * Story #4914 — explicit stdout ceiling for every git read in this module.
+ *
+ * `child_process.spawnSync` defaults `maxBuffer` to 1 MB. A committed
+ * baseline legitimately grows past that (a real consumer's crap baseline
+ * measured 1,178,910 bytes), at which point the child is killed —
+ * `status: null`, `signal: 'SIGTERM'`, `error.code: 'ENOBUFS'` — and the
+ * read fails for a reason that has nothing to do with the repository.
+ *
+ * 64 MB is not a new number: it is the bound already used at
+ * `run-test-profile.js:87`, `audit-baselines/trend.js:61` and
+ * `audit-baselines/weights.js:65`. This module was the outlier.
+ */
+const MAX_BUFFER_BYTES = 64 * 1024 * 1024;
 
 let _spawnSync = spawnSync;
 let _cache = new Map();
@@ -165,6 +185,7 @@ export function readBaseFromGit(ref, file, opts = {}) {
     encoding: 'utf-8',
     shell: false,
     env: cleanGitEnv(),
+    maxBuffer: MAX_BUFFER_BYTES,
   });
 
   // `child_process.spawnSync` returns `status: null` when the child died
@@ -232,6 +253,7 @@ export function readRangeSubjectsTouchingFile(baseRef, file, opts = {}) {
         encoding: 'utf-8',
         shell: false,
         env: cleanGitEnv(),
+        maxBuffer: MAX_BUFFER_BYTES,
       },
     );
   } catch {
--- a/.agents/scripts/lib/orchestration/check-baselines/phases/compare.js
+++ b/.agents/scripts/lib/orchestration/check-baselines/phases/compare.js
@@ -9,6 +9,7 @@
  * @module lib/orchestration/check-baselines/phases/compare
  */
 
+import { EXIT_CONFIG } from '../../../baselines/exit-codes.js';
 import { readBaseFromGit } from '../../../baselines/git-base.js';
 import { getKindModule } from '../../../baselines/kernel.js';
 import { resolveScope } from '../../../baselines/scope.js';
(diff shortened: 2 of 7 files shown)


dsj1984:
## 🔬 Code Review — Story #4914

**Comparison**: `main` … `story-4914`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

#### 🟢 Lint check passed with 2 warning(s) _[lint]_

Scoped lint reported 2 warning(s) on the changed surface. Treat as suggestions.

PR #4918, opened 2026-08-02
fix: stop the friction interceptor recording a signal about its own failure to spawn, and make the always-on invocation unambiguous (#4915)

### LINKED ISSUE #4915: Stop the friction interceptor recording a signal about its own failure to spawn, and make the always-on invocation unambiguous
## Goal
Make an agent that follows the always-on friction instruction verbatim actually record the friction it meant to record, instead of a row describing the interceptor's own ENOENT.

## Spec
`.agents/scripts/diagnose-friction.js` is a wrapper that executes the command it is given: `parseArguments` collects `cmdArgs` as the remaining argv words and `main()` calls `spawnSync(cmdArgs[0], cmdArgs.slice(1))`. The flag therefore requires argv-split words, but the always-loaded instruction in `.agents/instructions.md` writes it as a single `[FAILED_COMMAND]` placeholder, which invites quoting the whole command as one shell word. Quoted, the entire string becomes the executable name and the spawn fails ENOENT.

The contract to establish: a failure to *start* the child because the caller quoted the command is a **usage error in the interceptor's own invocation** — it must be reported as one and must write no ledger row. Today it is recorded as an `Execution Error` friction signal with `killOrigin: 'spawn-failure'`, which both discards the real friction and, on recurrence, makes the epilogue roll-up auto-file a framework-gap ticket about the framework's own instrumentation misuse.

One distinction is load-bearing and easy to get wrong: a genuinely missing executable and the quoted-string misuse are **indistinguishable from the `spawnSync` result** — both yield `status: null` with `ENOENT`. The discriminator lives in `cmdArgs`, not in the result. A correctly argv-split command whose binary is absent is real friction and MUST keep emitting the row it emits today; only the single-argument-containing-whitespace shape becomes a usage error.

A second shape in the same function is misdiagnosed and must be separated while it is open. `describeAbnormalExit` treats `signal === 'SIGTERM'` as proof of the interceptor's own `executionTimeoutMs` bound and advises splitting the command up. But a `maxBuffer` overflow presents identically — measured: `status: null`, `signal: 'SIGTERM'`, `error.code: 'ENOBUFS'` — so a command whose output exceeds the interceptor's 10 MiB `executionMaxBuffer` is reported as a timeout and the agent is told to shorten it, when the fix is a larger buffer. Story #4914 measured the same shape independently in `.agents/scripts/lib/baselines/git-base.js`. `error.code` is present on the result, so SIGTERM-with-ENOBUFS and SIGTERM-without-it are distinguishable and MUST be reported differently.

The always-loaded closure is under a byte ceiling with a context-budget ratchet, so the wording fix must not grow it.

## Changes
- `.agents/scripts/diagnose-friction.js` — refactors-existing
- `.agents/instructions.md` — refactors-existing
- `.agents/skills/core/diagnose-friction/SKILL.md` — refactors-existing
- `tests/diagnose-friction.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: `.agents/instructions.md` § 1.H writes the flag with a placeholder that reads unambiguously as multiple argv words (for example `--cmd <cmd> <args...>`), and the always-loaded closure's measured byte total does not increase relative to the pre-change value.
- [ ] AC-2: Invoking the interceptor with the whole command as one quoted argument — `cmdArgs.length === 1` and that argument contains whitespace — exits non-zero with a usage error that names the quoting mistake and shows the argv-split form, and appends NO row to the signals ledger.
- [ ] AC-3: A correctly argv-split command whose executable genuinely does not exist still appends an `Execution Error` row carrying `killOrigin: 'spawn-failure'` — the missing-tool friction signal is preserved, proving the fix discriminates on `cmdArgs` rather than on the ENOENT result.
- [ ] AC-4: A command that starts and exits non-zero, one killed by SIGTERM at the interceptor's `executionTimeoutMs` bound with no `ENOBUFS` error code (`killOrigin: 'interceptor-timeout'`), and one killed by another signal (`killOrigin: 'external'`) each still produce the category, `details`, and exit code they produce today.
- [ ] AC-5: `.agents/skills/core/diagnose-friction/SKILL.md` keeps its already-correct `--cmd <command args...>` form and states the quoted-string failure mode so the on-demand tier and the always-on tier no longer disagree.
- [ ] AC-6: No test in this change appends to the real friction ledger — each drives the CLI seam against an isolated temp root — and `describeAbnormalExit` remains module-local with no test-only export.
- [ ] AC-7: A child killed by SIGTERM whose result carries `error.code: 'ENOBUFS'` is reported as a buffer overflow, not a timeout: it gets its own `killOrigin` distinct from `interceptor-timeout`, and remediation naming the `executionMaxBuffer` bound (10 MiB) rather than advising the command be split into smaller steps.
- [ ] AC-8: A unit test drives both SIGTERM shapes through the CLI seam — one with `error.code: 'ENOBUFS'` and one without — and asserts they yield different `killOrigin` values and different remediation text, so the two cannot silently re-merge.
- [ ] AC-9: `npm test` and `npm run lint` both pass, including the pre-existing `diagnose-friction` and skill-doc assertions.

## Verify
- npm test (unit)
- npm run lint (validate)

## Non-Goals
- Changing what counts as friction for commands that actually start. A non-zero exit, a genuine interceptor timeout (SIGTERM without `ENOBUFS`), and an external kill must keep emitting exactly the rows they emit today. The one deliberate exception is SIGTERM carrying `error.code: 'ENOBUFS'`, which is a buffer overflow misreported as a timeout today.
- Raising the interceptor's `executionMaxBuffer` itself, or auditing other `spawnSync` sites for a missing `maxBuffer`. This Story makes the overflow *diagnosable*; #4914 raises the bound in `git-base.js`.
- Exporting `describeAbnormalExit` for tests. Its docblock states the export would fail CI's --production dead-exports ratchet; the CLI contract stays the test seam.
- Auto-recovering by shell-splitting a quoted command string. Guessing at shell quoting is worse than a loud usage error.
- Narrowing the follow-up roll-up's cross-run recurrence window, which is deliberate.
- The source-classifier basename gap, which is its own Story.

dsj1984:
Closes #4915

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Low Risk**
> Changes are confined to friction telemetry CLI behavior, docs, and tests; no auth, data, or production runtime paths.
> 
> **Overview**
> **`diagnose-friction.js`** no longer treats interceptor misuse as friction and labels **`maxBuffer`** kills separately from timeouts.
> 
> Callers who pass **`--cmd`** as one quoted string (e.g. `--cmd "npm run lint"`) get a **usage error** before spawn, with **no** `signals.ndjson` row—so ENOENT from a bogus executable name is not logged as wrapped-command friction. A correctly split argv with a missing binary still records **`spawn-failure`** friction.
> 
> When a child dies with **`SIGTERM`** and **`status: null`**, abnormal exits now branch on **`error.code === 'ENOBUFS'`**: **`killOrigin: buffer-overflow`**, category **Tool Limitation**, remediation that cites **`executionMaxBuffer`** (10 MiB) and explicitly says **not** to split the command; real interceptor timeouts stay **`interceptor-timeout`** with the existing split-or-raise-timeout advice.
> 
> Agent docs (**`instructions.md`**, **`diagnose-friction` skill**, CLI help) document argv-split **`--cmd`**. Tests cover quoting refusal, ledger skip, missing-binary discrimination, and overflow vs timeout remediation/details.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec693eca782ee56384304efbfd62f430ef1918a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/instructions.md
+++ b/.agents/instructions.md
@@ -79,7 +79,7 @@ present, else the **Tech Stack** section of `docs/architecture.md`.
 
 You MUST log operational friction (repeated tool errors, unrecoverable
 command failures, self-corrected ambiguity, automatable boilerplate):
-`node .agents/scripts/diagnose-friction.js --story [STORY_ID] --cmd [FAILED_COMMAND]`
+`node .agents/scripts/diagnose-friction.js --story [STORY_ID] --cmd <cmd> <args...>`
 — a **local NDJSON signal** (not posted to the ticket). Schema and the
 `AGENT_LOG_LEVEL` table:
 [`docs/execution-reference.md`](docs/execution-reference.md#friction-telemetry).
--- a/.agents/scripts/diagnose-friction.js
+++ b/.agents/scripts/diagnose-friction.js
@@ -16,7 +16,11 @@
  *
  * Usage:
  *   node diagnose-friction.js [--story <STORY_ID>] \
- *     [--epic <EPIC_ID>] --cmd <command with args...>
+ *     [--epic <EPIC_ID>] --cmd <cmd> <args...>
+ *
+ * `--cmd` consumes the remaining argv as separate words and spawns them with
+ * no shell. Quoting the whole command as one string is a usage error, not
+ * friction — it is refused loudly and writes no ledger row.
  *
  * Story/Epic resolution order:
  *   1. CLI flags (--story, --epic).
@@ -108,6 +112,16 @@ function classifyFrictionCategory(errorOutput) {
  */
 const INTERCEPTOR_TIMEOUT_SIGNAL = 'SIGTERM';
 
+/**
+ * A `maxBuffer` overflow presents identically to a timeout — `status: null`,
+ * `signal: 'SIGTERM'` — because Node kills the child the same way. The only
+ * discriminator is `result.error.code`, so a SIGTERM carrying this code is a
+ * buffer overflow and must never be reported as a timeout (Story #4915).
+ *
+ * @type {string}
+ */
+const OVERFLOW_ERROR_CODE = 'ENOBUFS';
+
 /** Shell convention for "the process died by signal N": exit `128 + N`. */
 const SIGNAL_EXIT_BASE = 128;
 
@@ -118,8 +132,10 @@ const SIGNAL_EXIT_BASE = 128;
  * interceptor would report success for a command it just watched get killed
  * (Story #4851).
  *
- * Which signal fired is the diagnostic value: SIGTERM points at the
- * interceptor's own bound, anything else at the host. Recording that plus the
+ * Which signal fired is the diagnostic value: SIGTERM points at one of the
+ * interceptor's own bounds, anything else at the host. A SIGTERM splits again
+ * on `error.code`: `ENOBUFS` means the output blew past `executionMaxBuffer`,
+ * anything else means `executionTimeoutMs` fired. Recording that plus the
  * bound itself is what makes the row actionable to a consumer who cannot edit
  * the materialized framework tree.
  *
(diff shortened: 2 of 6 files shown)


dsj1984:
## 🔬 Code Review — Story #4915

**Comparison**: `main` … `story-4915`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4920, opened 2026-08-02
fix: classify a framework script named by bare basename as framework rather than consumer (#4916)

### LINKED ISSUE #4916: Classify a framework script named by bare basename as framework rather than consumer
## Goal
Stop a friction signal being routed as consumer-actionable — and discarded — merely because its command named a framework script by basename instead of by path.

## Spec
`classifySignalSource` in `.agents/scripts/lib/observability/source-classifier.js` decides `framework` vs `consumer` by scanning `failingPath` and `emitter.command` for a substring from `FRAMEWORK_PREFIXES` (`.agents/`, `.agentrc.json`, `.claude/`, `node .agents/scripts/`). A framework script referenced by bare basename contains none of them, so it classifies `consumer`.

Measured on the live function: `node .agents/scripts/acceptance-eval.js --story 4901` yields `framework`, while `acceptance-eval.js --story 4901` and `single-story-close.js --story 4906` both yield `consumer`. Two signals of otherwise identical shape can therefore carry different `source` values, and a real framework defect classified `consumer` is routed as consumer-actionable and discarded — the failure class Stories #4824 and #4825 addressed by a different mechanism.

The contract to establish: recognition MUST NOT depend on how the caller spelled the reference. A bare basename that names a script shipped under `.agents/scripts/` classifies `framework`.

Two invariants bound the fix. The classifier is called per signal and MUST stay pure — no filesystem I/O at classify time — so the recognized basename set is static data, guarded by a test that fails when a script is added or renamed without updating it; a silently stale set would rebuild the same blind spot. And `framework`-wins precedence plus every existing prefix behaviour MUST be preserved: widening recognition may only ever move a signal from `consumer` to `framework`, never the reverse.

## Changes
- `.agents/scripts/lib/observability/source-classifier.js` — refactors-existing
- `tests/lib/observability/source-classifier.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: AC-1: `classifySignalSource` returns `framework` for a signal whose `emitter.command` names a framework script by bare basename with no path segment — asserted for at least `single-story-close.js --story 4906` and `acceptance-eval.js --story 4901`, the two live shapes measured on current main.
- [ ] AC-2: AC-2: A command carrying no framework reference still returns `consumer` — asserted for a plain tool invocation such as `npm test` and for a script basename that is not shipped under `.agents/scripts/`, so the widening does not collapse the classifier into always-framework.
- [ ] AC-3: AC-3: A test asserts the recognized basename set matches the actual set of scripts shipped under `.agents/scripts/`, and fails when a script is added, renamed, or removed without the set being updated — so the set cannot go stale and silently restore the blind spot.
- [ ] AC-4: AC-4: Every classification that returns `framework` today still returns `framework`: each of the four `FRAMEWORK_PREFIXES` forms, the `failingPath` route, the details-scan tier, and the tool-degraded category tier are asserted unchanged, and no input that classified `framework` before now classifies `consumer`.
- [ ] AC-5: AC-5: `classifySignalSource` performs no filesystem access when classifying — the basename set is resolved as static data, asserted by a test that would fail if a read happened per call.
- [ ] AC-6: AC-6: `npm test` and `npm run lint` both pass, with the pre-existing `source-classifier` assertions unchanged.

## Verify
- npm test (unit)
- npm run lint (validate)

## Non-Goals
- Filesystem I/O at classify time. The classifier is called per signal and stays pure; the basename set is static data kept honest by a sync test.
- Changing FRAMEWORK_PREFIXES, the framework-wins precedence, or the details/category fallback tiers of classifySignalSource.
- Reclassifying anything from framework to consumer. This widens recognition in one direction only.
- The diagnose-friction invocation contract and its spawn-failure signal, which is its own Story.

dsj1984:
Closes #4916

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Low Risk**
> Additive routing heuristic in observability tagging with broad unit coverage; no auth, data, or runtime I/O changes.
> 
> **Overview**
> Friction signals that blame a Mandrel CLI by **bare script name** (e.g. `acceptance-eval.js --story 4901`) were tagged `consumer` while the same invocation with a full `.agents/scripts/` path was `framework`, so real framework failures could be dropped from the framework retro stream.
> 
> **`source-classifier.js`** adds a static list of top-level `.agents/scripts/*.js` basenames and treats whitespace-delimited command tokens (with light quote/comma trimming) as framework when they match that set. Path-prefix rules are unchanged; matching is routed through `namesFrameworkSurface`, which is **additive** so nothing should flip from `framework` to `consumer`. Library paths under `lib/` are intentionally excluded to avoid false positives on generic names like `notify.js`.
> 
> **Tests** cover the production command shapes, pathed vs bare parity for every listed script, guardrails (consumer paths, substring mismatches), regression on prefix/details/`tool-degraded` tiers, a directory sync check so the basename list cannot drift, and asserts the classifier stays pure (no `fs` at classify time).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f2f6a4687da6d658acaf6ccd304deee7129d2c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/observability/source-classifier.js
+++ b/.agents/scripts/lib/observability/source-classifier.js
@@ -17,6 +17,9 @@
  *       - `.agentrc.json`
  *       - `.claude/`
  *       - `node .agents/scripts/`
+ *     …or (Story #4916) names one of the CLI entry points shipped at the
+ *     top level of `.agents/scripts/` by **bare basename**, so recognition
+ *     does not depend on how the caller spelled the reference.
  *   - Anything else (or empty input) defaults to `"consumer"`. The default
  *     is intentional — most friction comes from consumer code touching
  *     framework tooling, and we'd rather under-tag than mis-route a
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4916

**Comparison**: `main` … `story-4916`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4939, opened 2026-08-02
fix: carry audit provenance footers onto dependent Stories so the next sweep can dedup against them (#4935)

### LINKED ISSUE #4935: Carry audit provenance footers onto dependent Stories so the next sweep can dedup against them
## Goal
A Story with a `depends_on` edge persists without its `audit-fingerprints` / `audit-semantic-keys` footers, so the next audit sweep re-files work it already planned.

## Spec
`assembleOnePlanStory` calls `carryProvenanceFooters({ from: opts.provenanceSource, into: serialized })` and appends the footers to the **serialized body string**. The `blocked by #<id>` footer cannot be written until the sibling's real issue number exists, so the dependency step re-serializes the body from `bodyObject` — which never held the carried footers. The carry is silently lost for exactly the Stories that carry an edge.

Observed live on `plan-run::6cc05d2b`: #4922, #4924 and #4925 each persisted with the full 78-sha union; #4923 and #4926, the only two with `depends_on`, persisted with **zero** footers. Confirmed by inspecting the issue bodies after the run.

The consequence is not cosmetic. `/audit-to-stories` Phase 6 confirms a search hit by parsing the canonical footer marker out of an issue body; a footerless Story is invisible to that confirmation, so the finding routes `new` and a duplicate Story is opened. A plan in which every Story carries an edge loses provenance for the entire batch.

The fix must make the carry survive whatever re-serialization the dependency step performs — either by carrying the footers on `bodyObject` so any re-serialize preserves them, or by re-applying the carry after the blocked-by footer is written. `carryProvenanceFooters` is already additive, union-preserving and idempotent, so re-applying is safe by construction; do not weaken those properties.

Do **not** work around this by having the authoring agent hand-copy footers. The workflow states the carry is mechanical on purpose, and a remembered step is no step at all.

The regression test MUST cover the dependent case specifically — a Story with a non-empty `depends_on` persisted from a seed carrying footers must end with both footers AND its `blocked by #<id>` line. CORRECTED DURING DELIVERY: the loss site is `renderStoryBodyForCreate` (~line 626), which `createStoryIssues` calls once a sibling's issue number is known — not the `assembleOnePlanStory` call site named above; the re-serialization mechanism was right, the function was not. Also corrected: `route-finding.contract.test.js` is the `searchIssues` port contract and never imports plan-persist, so the regression belongs in `plan-persist-story-ops.test.js` where the dependent path is reachable.

Re-stamping the two Stories that already landed footerless (#4923, #4926) is out of scope; they are closed, and the union they would have carried is already present on their three siblings.

## Changes
- `.agents/scripts/lib/orchestration/plan-persist/story-ops.js` — refactors-existing
- `.agents/scripts/lib/findings/route-finding.js` — refactors-existing
- `tests/lib/findings/route-finding.contract.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: A Story authored with a non-empty `depends_on` and persisted from a seed carrying `audit-fingerprints` / `audit-semantic-keys` footers ends with BOTH footers in its body.
- [ ] AC-2: That same Story still carries its `blocked by #<id>` line, and the two coexist.
- [ ] AC-3: A regression test exercises the dependent path explicitly and fails against the pre-fix implementation.
- [ ] AC-4: `carryProvenanceFooters` remains additive, union-preserving and idempotent — applying it twice to the same body adds no duplicate sha.
- [ ] AC-5: The fingerprint of a persisted Story is unchanged by this fix (no re-minting): the existing `route-finding.contract.test.js` assertions still pass.
- [ ] AC-6: A Story with no `depends_on` is unaffected — it carries the same footers it did before.

## Verify
- node --test tests/lib/findings/route-finding.contract.test.js (contract)
- npm run lint (validate)

## Non-Goals
- Re-stamping the already-landed footerless Stories #4923 and #4926.
- Changing the fingerprint algorithm — `severity` and `labels` are identity fields folded into the sha, and re-minting would break dedup for every finding already filed.

dsj1984:
Closes #4935

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Low Risk**
> Narrow change to posted issue bodies for dependent Stories during plan persist, with regression tests; no auth or broad orchestration behavior changes beyond audit dedup metadata.
> 
> **Overview**
> Fixes a bug where **Stories with `depends_on` siblings** were persisted **without** `audit-fingerprints` / `audit-semantic-keys` footers, so `/audit-to-stories` could not confirm them and might open duplicate Stories.
> 
> At create time, `renderStoryBodyForCreate` must re-serialize the body from `bodyObject` to turn slug dependencies into `blocked by #N` lines. That path replaced the assembled `story.body` string and **dropped** audit footers that assembly had only appended to the string, not to `bodyObject`. The change runs **`carryProvenanceFooters`** from `story.body` into the re-serialized markdown before appending the plan fingerprint marker; stories with no dependencies are unchanged.
> 
> Tests now exercise **`createStoryIssues`** (not just assembly) for a dependent cohort: footers survive alongside the blocked-by footer, `parseFingerprintFooter` matches Phase 6 confirmation, and footers are not duplicated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df5fdbdf53fb4ee96f7cbce4afe852937c4771ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/orchestration/plan-persist/story-ops.js
+++ b/.agents/scripts/lib/orchestration/plan-persist/story-ops.js
@@ -604,6 +604,21 @@ function warnOnDivergentSameTitleStory(story, idsByTitle) {
  * sibling `depends_on` slugs resolved to real issue ids, plus the invisible
  * plan-fingerprint marker that makes the create loop resumable.
  *
+ * **The dependent branch re-serializes, so it must re-carry the provenance**
+ * (Story #4935). `assembleOnePlanStory` appends the audit footers to the
+ * serialized body *string*, but a `blocked by #N` footer cannot be written
+ * until the sibling's real issue number exists — so this function rebuilds the
+ * markdown from `bodyObject`, which never held those footers. Without the
+ * re-carry, provenance was silently dropped for exactly the Stories carrying an
+ * edge (observed live on `plan-run::6cc05d2b`: the two Stories with a
+ * `depends_on` persisted with zero footers while their three siblings carried
+ * the full union), which makes them invisible to `/audit-to-stories` Phase 6's
+ * footer confirmation and re-files work already planned. `story.body` is the
+ * carried assembly, so it is the provenance source here;
+ * `carryProvenanceFooters` is additive, union-preserving and idempotent, so
+ * re-applying it is safe by construction. The fingerprint is computed upstream
+ * over `story.body` and is untouched by this — no re-minting.
+ *
  * @param {object} story
  * @param {Map<string, number>} idBySlug
  * @returns {string}
@@ -612,13 +627,17 @@ function renderStoryBodyForCreate(story, idBySlug) {
   const dependencyRefs = story.depends_on.map(
     (slug) => `#${idBySlug.get(slug)}`,
   );
-  const base =
-    dependencyRefs.length === 0
-      ? story.body
-      : serializeStoryBody(
-          { ...story.bodyObject, depends_on: dependencyRefs },
-          { includeFooter: true },
-        );
+  let base = story.body;
+  if (dependencyRefs.length > 0) {
+    const reserialized = serializeStoryBody(
+      { ...story.bodyObject, depends_on: dependencyRefs },
+      { includeFooter: true },
+    );
+    base = carryProvenanceFooters({
+      from: story.body,
+      into: reserialized,
+    }).body;
+  }
   return `${base}\n\n${planFingerprintMarker(story.fingerprint)}`;
 }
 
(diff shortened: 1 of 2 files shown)


dsj1984:
## 🔬 Code Review — Story #4935

**Comparison**: `main` … `story-4935`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4940, opened 2026-08-02
fix: read prior baseline rows per kind so duplication's diff-scope preservation and epsilon damping actually work (#4937)

### LINKED ISSUE #4937: Read prior baseline rows per kind so duplication's diff-scope preservation and epsilon damping actually work
## Goal
`readPriorBaselineRows` routes every non-`crap` kind through a reader that filters on a field only maintainability rows carry, so duplication always sees an empty prior.

## Spec
In `lib/baselines/diff-scope-cli.js`, `readPriorBaselineRows` dispatches `crap` to `readCrapPriorRows` and **everything else** to `readMaintainabilityPriorRows`, which keeps only rows satisfying `typeof r.mi === 'number'`. Duplication rows are keyed on `percentage`, not `mi`, so every row is filtered out and `update-duplication-baseline.js` always receives an empty `prior`.

Two consequences, both silent:

- **Epsilon damping is inert for duplication.** Damping compares a new reading against the prior row to suppress sub-threshold churn. With no prior, every reading is treated as new and nothing is damped.
- **`--diff-scope` corrupts out-of-scope rows.** Preservation works by carrying prior rows forward for files outside the diff. With an empty prior there is nothing to carry. CORRECTED DURING DELIVERY: the symptom is not truncation — `update-duplication-baseline.js` always runs a FULL jscpd scan, so out-of-scope rows are silently OVERWRITTEN by the rescan, and any file the rescan no longer reports is dropped outright. Reproduced pre-fix: prior length 0, rows written [[a,40],[b,0],[c,0]] — `src/b.js` fell from 30 to 0. Also corrected: the blast radius is SIX of the eight kinds, not duplication alone; `lighthouse` (keyed on `route`) and `bundle-size` (keyed on `bundle`) also failed the reader's `typeof r.path === 'string'` clause before `mi` was consulted.

The second is currently masked: recent duplication regenerations have been full-scope rewrites, where an empty prior is harmless. The defect is latent, not benign — the first diff-scoped duplication run destroys the baseline.

The reader must select per kind rather than defaulting non-`crap` to the maintainability shape. Check the other kinds against their real row schemas rather than assuming only duplication differs — the same default has been applied to every non-`crap` kind, so any kind whose rows lack `mi` has the same bug.

The regression test MUST prove a diff-scoped duplication run **preserves** out-of-scope rows. A test that only asserts the prior is non-empty would pass on a reader that still returns the wrong shape.

## Changes
- `.agents/scripts/lib/baselines/diff-scope-cli.js` — refactors-existing
- `.agents/scripts/update-duplication-baseline.js` — refactors-existing
- `tests/baselines/writer-scope.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: `readPriorBaselineRows` selects a reader per kind; no kind falls through to a reader that filters on a field its rows do not carry.
- [ ] AC-2: `update-duplication-baseline.js` receives a non-empty `prior` when `baselines/duplication.json` has rows.
- [ ] AC-3: A test proves a `--diff-scope` duplication run PRESERVES rows for files outside the diff, rather than only asserting the prior is non-empty.
- [ ] AC-4: Every baseline kind whose rows lack an `mi` field is covered — the audit is not limited to duplication.
- [ ] AC-5: A test passes an explicit epsilon (e.g. 0.5) to the duplication write path and asserts a reading whose delta is below it folds back to the prior `percentage` rather than being written as new.
- [ ] AC-6: `node .agents/scripts/check-baselines.js --gate duplication` still passes and no committed baseline is rewritten by this Story.

## Verify
- node --test tests/baselines/writer-scope.test.js (unit)
- node --test tests/baselines/refresh-service.diff-scope.test.js (contract)
- node .agents/scripts/check-baselines.js --gate duplication (validate)
- npm run lint (validate)

## Non-Goals
- Regenerating any committed baseline — this fixes the reader, not the data.
- Changing epsilon-damping thresholds or the `--diff-scope` resolution logic itself.

dsj1984:
Closes #4937

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes baseline write behavior for manual update CLIs using `--diff-scope`; incorrect contracts could still corrupt baselines, but scope is limited to the diff-scope helper with strong regression tests.
> 
> **Overview**
> Fixes **`--diff-scope`** baseline updates for non-maintainability kinds (especially **duplication**), where the prior reader wrongly applied a maintainability (`path` + `mi`) filter so priors came back empty and scoped writes could **truncate** baselines and skip **epsilon** damping.
> 
> **`readPriorBaselineRows`** now validates rows per kind via **`PRIOR_ROW_METRIC_FIELDS`** and each kind’s **`keyField`** from **`getKindModule`**, throws on unknown kinds instead of silently emptying the prior, and **`buildWriterScopeArgs`** accepts injectable **`fsImpl`** / **`spawnImpl`** for hermetic tests. **`baselines/dead-exports.json`** drops the unused **`readPriorBaselineRows`** export entry.
> 
> **`writer-scope.test.js`** adds Story #4937 coverage: all kinds round-trip canonical rows, duplication **`--diff-scope`** preserves out-of-scope rows (including rows missing from a rescan), and sub-epsilon duplication metrics fold to the prior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb8480a40b08054bfd2bc8203669ec5bc97e8c5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/scripts/lib/baselines/diff-scope-cli.js
+++ b/.agents/scripts/lib/baselines/diff-scope-cli.js
@@ -170,12 +170,14 @@ const PRIOR_ROW_METRIC_FIELDS = Object.freeze({
  * every unlisted kind the maintainability filter, so an unmodelled kind
  * read as "prior is empty" instead of "prior is unreadable".
  *
- * Pure; no I/O.
+ * Pure; no I/O. Module-private: `readPriorBaselineRows` is the only caller
+ * and the only surface tests drive it through, so exporting it would add an
+ * unused export the dead-export ratchet refuses.
  *
  * @param {string} kind
  * @returns {(row: object) => boolean}
  */
-export function priorRowPredicateFor(kind) {
+function priorRowPredicateFor(kind) {
   const metricFields = PRIOR_ROW_METRIC_FIELDS[kind];
   if (!metricFields) {
     throw new Error(
--- a/.agents/scripts/lib/baselines/diff-scope-cli.js
+++ b/.agents/scripts/lib/baselines/diff-scope-cli.js
@@ -17,11 +17,18 @@
  * four scripts: same argv parser, same git invocation, same forward-slash
  * path normalisation. The four CLIs differ only in how they pipe the
  * resolved scope through to their writer.
+ *
+ * Preservation is only as good as the prior it reads, so the prior reader
+ * dispatches on the kind's own row contract (see `PRIOR_ROW_METRIC_FIELDS`)
+ * rather than defaulting every non-`crap` kind to the maintainability row
+ * shape — that default filtered every duplication row out and left the
+ * writer with nothing to preserve (Story #4937).
  */
 
 import { spawnSync } from 'node:child_process';
 import fs from 'node:fs';
 import { parseNameOnlyStdout } from '../changed-files.js';
+import { getKindModule } from './kernel.js';
 
 /**
  * Parse `--diff-scope <ref>` (and the legacy `--diff-scope=<ref>` form)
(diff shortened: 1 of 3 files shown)


dsj1984:
## 🔬 Code Review — Story #4937

**Comparison**: `main` … `story-4937`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 1

### 🟢 Suggestion (1)

PR #4942, opened 2026-08-02
fix: close three harness gaps: unbounded CI jobs, divergent runner flags, and inert pretest hooks (#4936)

### LINKED ISSUE #4936: Close three harness gaps: unbounded CI jobs, divergent runner flags, and inert pretest hooks
## Goal
Three parts of the test and CI harness silently do not run what they appear to run, and each was found only by an incident rather than by a check.

## Spec
Three independent gaps, one subsystem — the harness that everything else's correctness claims rest on.

**No job timeout.** Zero of the four files under `.github/workflows/` set `timeout-minutes`, so every job inherits GitHub's 360-minute default. A deadlocked `Windows Smoke` job burned 44 minutes of a `windows-latest` runner and would have burned six hours; worse, because the run stayed `in_progress`, GitHub withheld logs for the *required* check that had already failed on the same run, so the actual blocker could not be diagnosed until the run was cancelled by hand. Set a per-job `timeout-minutes` sized to each job's observed duration with headroom — do not pick one global number that is generous enough to be useless.

**Divergent runner flags.** `run-tests.js:75` passes `--experimental-test-module-mocks`; `run-coverage.js` does not. Any test using `t.mock.module` therefore passes under `npm test` and fails under `npm run test:coverage` — which is the *required* CI job. The two runners must not diverge on flags that decide whether a test can run at all; make the flag set a single shared value both consume, so a future flag cannot drift again.

**Inert lifecycle hooks.** `.npmrc` sets `ignore-scripts=true` as deliberate defence against malicious postinstall hooks (CWE-1357), and that setting is correct and stays. But it also suppresses npm `pre*` hooks, so `pretest`, `pretest:quick` and `pretest:integration` never fire for any tier. Story #4922 worked around this for one tier by having CI invoke `pretest:coverage` by name. Either every tier invokes its preflight explicitly the same way, or the hooks are removed so nothing appears to run a preflight that does not. A script that looks like a gate and never executes is worse than no script.

Each gap needs a check that would have caught it, not just a fix — a workflow-lint assertion for the timeout, a test pinning the two runners' flag sets equal, and a test that the preflight actually executes for each tier.

## Changes
- `.github/workflows/ci.yml` — refactors-existing
- `.github/workflows/install-matrix.yml` — refactors-existing
- `.github/workflows/issue-body-conformance.yml` — refactors-existing
- `.github/workflows/release-please.yml` — refactors-existing
- `.agents/scripts/run-coverage.js` — refactors-existing
- `.agents/scripts/run-tests.js` — refactors-existing
- `.agents/scripts/lib/test-tiers.js` — refactors-existing
- `package.json` — refactors-existing
- `tests/scripts/run-coverage.test.js` — refactors-existing
- `tests/lib/test-tiers.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: Every job in every file under `.github/workflows/` sets an explicit `timeout-minutes` between 10 and 30, sized per job from its observed duration (currently Validate and Test 155s, Windows Smoke 141s, baselines 19s) rather than one global value.
- [ ] AC-2: A check enumerates every `jobs.<id>` key across `.github/workflows/*.yml` and exits non-zero when any lacks `timeout-minutes` or declares one above 30.
- [ ] AC-3: `run-coverage.js` and `run-tests.js` derive their node flag set from one shared value; neither restates a flag literal the other lacks.
- [ ] AC-4: A test asserts the two runners' flag sets are equal and fails if a flag is added to one only.
- [ ] AC-5: A test using `t.mock.module` passes under BOTH `npm test` and `npm run test:coverage`.
- [ ] AC-6: Each test tier either invokes its preflight explicitly, or has no `pretest*` script — no script exists that appears to be a preflight and never executes.
- [ ] AC-7: A test proves the preflight actually runs for each tier that declares one.
- [ ] AC-8: `.npmrc` still sets `ignore-scripts=true`.

## Verify
- node --test tests/scripts/run-coverage.test.js (unit)
- node --test tests/lib/test-tiers.test.js (unit)
- npm run test:coverage (validate)
- npm run lint (validate)

## Non-Goals
- Removing or relaxing `ignore-scripts=true` in `.npmrc` — it is deliberate security hardening and stays.
- Retuning any coverage floor or baseline.

dsj1984:
Closes #4936

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Touches required CI paths (coverage runner, all workflows) and plan-persist body rendering for dependent stories without the former provenance re-carry tests—worth confirming the story-ops change is intentional.
> 
> **Overview**
> Closes **Story #4936** by tightening CI harness behavior in three places: job timeouts, shared test-runner flags/preflight, and lint enforcement.
> 
> **GitHub Actions timeouts** — Adds `check-workflow-timeouts.js` (10–30 minute window per job, YAML-aware, reusable-workflow jobs exempt) and wires it into `run-lint.js`. All workflow jobs get explicit `timeout-minutes` sized from observed runtimes (e.g. **Windows Smoke** at 15 minutes).
> 
> **Test runner contract** — New `lib/test-runner-contract.js` centralizes `TEST_RUNNER_FLAGS` (including `--experimental-test-module-mocks` and host-clamped `--test-concurrency`) and `runTierPreflight`. `run-tests.js` and `run-coverage.js` call preflight before spawning tests; `runCoveragePipeline` is refactored for injectable stages and runs full-tier preflight first. Inert `pretest*` scripts are removed from `package.json`; CI no longer runs `npm run pretest:coverage` before coverage.
> 
> **Tests & baselines** — New suites for the timeout gate and runner contract; coverage tests assert flag parity between runners. Coverage/maintainability baselines refresh for new/changed scripts.
> 
> **Unrelated regression risk** — `renderStoryBodyForCreate` drops `carryProvenanceFooters` on the dependent-story re-serialize path and removes the Story #4935 persist tests — dependent Stories may lose audit provenance footers again unless that revert was intentional.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fa32f2262d833e29598fb06a518b04e14a52fad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4936

**Comparison**: `main` … `story-4936`
**Provider**: `native`
**Findings**: 0

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 0
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### ✅ No findings

PR #4970, opened 2026-08-03
fix: give anonymous CRAP methods an order-independent identity so the comparator stops pairing unrelated functions (#4969)

### LINKED ISSUE #4969: Give anonymous CRAP methods an order-independent identity so the comparator stops pairing unrelated functions
## Goal
Stop the CRAP gate reporting phantom regressions (and masking real ones) by replacing the ordinal-derived identity that 34.6% of baseline rows depend on, and refresh a file's rows when a Story edits it.

## Spec
**The defect.** `typhonjs-escomplex-commons` names unnamed functions `<anon method-N>` from a per-module counter incremented during the AST walk, so inserting or removing any anonymous function renumbers every later one. `.agents/scripts/lib/crap-engine.js` adopts that label verbatim as `method`, and `.agents/scripts/lib/baselines/kinds/crap.js` keys rows `path::method@startLine`. On an exact-key miss the comparator pairs candidates sharing `path::method` by nearest `startLine` (`pickDriftCandidate`) — correct for a named function that moved, unsound for an ordinal — then reports `drifted-regression`. The module already refuses to pair across coordinate systems for the same reason; ordinal identity has no such guard. **1463 of 4233 rows (34.6%) are anonymous.**

**Contract.** Anonymous methods get an identity that is a function of the method itself, not its position among siblings — an enclosing-named-scope path plus a within-scope discriminator, or a normalized AST/source-shape hash. Whatever is chosen must be stable when unrelated anonymous functions are added or removed elsewhere in the file, and must distinguish two genuinely different anonymous functions in the same scope. Named methods keep their current identity.

The schema (`additionalProperties: false`) admits the new field explicitly. `SCORING_SEMANTICS` (currently `coverage-join-v2`) **must** bump: the module's own contract is that rows scored under different identity rules are incomparable, and the bump is what stops old rows being paired with new ones during the migration. Re-keying rewrites most of `baselines/crap.json` — that is the intended migration, not the churn the repo warns about.

**Refresh path.** `update-crap-baseline` preserves out-of-scope rows verbatim and nothing refreshes a file the delivery actually edited, so coordinates rot until a mismatched pair happens to score worse. A file in the change set must have its rows re-scored. Remove the `refresh-service.js` workaround comment that tells authors to avoid arrow functions.

## Changes
- `.agents/scripts/lib/crap-engine.js` — refactors-existing
- `.agents/scripts/lib/baselines/refresh-service.js` — refactors-existing
- `.agents/schemas/baselines/crap.schema.json` — refactors-existing
- `baselines/crap.json` — refactors-existing
- `tests/lib/crap-engine.test.js` — refactors-existing
- `tests/check-crap-json.test.js` — refactors-existing

## Acceptance
- [ ] AC-1: An anonymous method's row identity is unchanged when an unrelated anonymous function is added or removed earlier in the same file, asserted by a test that performs exactly that edit and observes no regression reported
- [ ] AC-2: Two distinct anonymous functions within the same enclosing scope receive different identities, so the re-keying cannot collapse separate methods into one row
- [ ] AC-3: A genuine complexity increase in an anonymous function is still reported as a regression after the re-keying, proving the fix did not buy silence by dropping coverage
- [ ] AC-4: SCORING_SEMANTICS is bumped and rows carrying the previous value are treated as incomparable rather than pickDriftCandidate-paired, so the migration reports no phantom regressions
- [ ] AC-5: crap.schema.json admits the new identity field and baselines/crap.json validates against it with every row re-keyed
- [ ] AC-6: A file present in the refresh scope has its rows re-scored rather than preserved verbatim, and the refresh-service comment instructing authors to avoid arrow functions is gone
- [ ] AC-7: Appending a single comment line to .agents/scripts/check-test-temp-hygiene.js no longer produces a CRAP violation

## Verify
- npm test -- tests/lib/crap-engine.test.js (unit)
- npm test -- tests/check-crap-json.test.js (unit)
- node .agents/scripts/quality-preview.js --changed-since origin/main (validate)

## Non-Goals
- Changing crapFormula — the arithmetic is correct and coverage is already handled as a [0,1] fraction
- Weakening, raising, or retuning the configured CRAP floors
- Removing the existing coordinate-system and provenance guards
- Patching typhonjs-escomplex-commons upstream — the label is re-keyed at our join layer
- Resolving anything with a --full-scope refresh outside the deliberate identity migration

dsj1984:
Closes #4969

_Auto-opened by `/deliver`._

---

> [!NOTE]
> **Medium Risk**
> Changes half of the CRAP row identity key for many anonymous methods and requires a full CRAP baseline re-seed; mitigated by explicit scoring-semantics and half-migration fail-closed checks.
> 
> **Overview**
> **Anonymous CRAP methods** are no longer keyed by escomplex’s position-dependent `<anon method-N>` labels. A new `crap-method-identity` module derives `<anon {enclosing-scope-path}>` identities (parameter list + sibling ordinal within a scope), and the scorer persists an optional row flag `anonymous: true` while named methods stay unchanged.
> 
> **Baseline contract** moves to `scoringSemantics: method-identity-v3` so old and new row keys are not compared across the boundary. A fatal **`anon-identity-unstamped`** compat axis catches half-migrated baselines (diff-scoped refresh leaving ordinal-keyed anonymous rows under a v3 envelope). Schema and projection/load paths now carry `anonymous` so that check cannot be undermined.
> 
> **Writer fix:** `provenanceStamped` (and related stamps) are always included in the on-disk envelope projection—fixing a gap where stamps validated in memory were dropped at write time.
> 
> Minor: duplication diff-scope predicate is inlined again (ordinal renumbering concern removed); `coverage.json` picks up the new identity module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd5db020acb20cae393eba8d1c7d8c7663c3c17f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dsj1984:
## 🔬 Code Review — Story #4969

**Comparison**: `main` … `story-4969`
**Provider**: `native`
**Findings**: 1

### 📦 Severity Tier Counts

- 🔴 Critical Blocker: 0
- 🟠 High Risk: 1
- 🟡 Medium Risk: 0
- 🟢 Suggestion: 0

### 🟠 High Risk (1)

#### 🟠 Lint check failed (1 error(s)) _[lint]_

Scoped lint reported 1 error(s) and 0 warning(s) on the changed surface. Fix errors before merging.

